PR #18, opened 2020-03-10
Fix missing route parameter type in code snippet for modifying blog post route

umar-ahmed:


cannikin:
Thanks for the PR @umar-ahmed  

We actually do *not* want to show the user `:Int` at this point in the tutorial. We show the default behavior to start and then we fix the route by adding `:Int` a little later in the text.

umar-ahmed:
@cannikin Yeah sorry, I realized that once I continued further in the tutorial :sweat_smile:. Nice docs btw!

cannikin:
Thanks! We've got a lot we want to add still but we hope this enough to get people excited about the framework!

PR #20, opened 2020-03-10
Fix typos in TUTORIAL.md

migueloller:


cannikin:
Thanks for catching and fixing those!

PR #24, opened 2020-03-12
Change DB_HOST env var to DATABASE_URL

peterp:
Prisma recommends that we use `DATABASE_URL` instead of `DB_HOST` env vars, this should only be merged after https://github.com/redwoodjs/create-redwood-app/pull/33 `(0.2.0)` has been released.


cannikin:
We'll want to update the example apps as well.

peterp:
I'll handle those tomorrow!

PR #30, opened 2020-03-13
Add style button hover

leonardoelias:
Hey adding style to the button state.

ezgif-3-1a320fa9f060


cannikin:
Great add, thanks!

PR #31, opened 2020-03-13
Rename posts.sdl.js to proper file name

jvanbaarsen:
The filename was not the same as actually being used.


thedavidprice:
Great catch. Thanks!

Issue #1, opened 2020-02-17
Launch website


cannikin:
Should contain a nice splash/home page and the full tutorial broken down into manageable sections, one page each.

* static site
* build script that downloads the README, splits it onto pages based on h2-level headers and creates a page for each

thedavidprice:
FWIW — this might be helpful for tutorial code sections:
https://clipboardjs.com

Have not taken it for a spin...

PR #32, opened 2020-03-14
Spread data to avoid HTTP status code 400

AryanJ-NYC:
The docs have a mistake that results in bad GraphQL input:
GraphQL error]: Message: Variable "$input" got invalid value { data: { name: "AH", email: "B@b.com", message: "fsdaf" } }; Field "data" is not defined by type ContactInput.

Screen Shot 2020-03-13 at 8 04 45 PM

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -1882,7 +1882,7 @@ const ContactPage = (props) => {
   const [create] = useMutation(CREATE_CONTACT)
 
   const onSubmit = (data) => {
-    create({ variables: { input: data }})
+    create({ variables: { input: ...data }})
     console.log(data)
   }
 


pantharshit00:
Hey, I checked this out and docs are correct. I suspect you might have wrapped the data object in your code inside an object by mistake like so:
```js
create({ variables: { input: { data } }})
```

That is why the error message that you have pasted is saying that it received an object with key `data` in it.

In the docs, it directly passed, which is correct:
```js
create({ variables: { input: data }})
```

Obviously, if you will spreadthe data object in snippet #1, it makes it equivalent to snippet #2

AryanJ-NYC:
D'oh. Definite egg on my face. You're right. I _did_ wrap my data object. 🤦‍♂️

thedavidprice:
@AryanJ-NYC thanks for opening the PR. Also, I flipped through your youtube video and saw you’d like to help with Jest and Eslint. Yes, please! Ping me over in issue https://github.com/redwoodjs/redwood/issues/181 and let’s organize and focus a bit.

@pantharshit00 thanks so much for the helpful triage! Really appreciate people getting involved right out of the gate.

PR #35, opened 2020-03-15
Destructure children in generated layout props

michelegera:
This matches the layout generator template as per https://git.io/JviHr.


cannikin:
Ahh yes, updated the generator but forgot about these code samples, thanks!

PR #37, opened 2020-03-15
Fix tutorial typo: Loader component → Loading

lachlanjc:
Pretty sure this is a typo in the tutorial, referring to the `Loading` component as `Loader`.


cannikin:
Yup good catch.

PR #61, opened 2020-03-22
tutorial conventions around files. clarity?

paulmars:
most of the examples are JS files or CLI inputs which can be copied, this breaks convention

looks like broken javascript if you have it stored in a js file, but is actually SDL. should be one or the other


cannikin:
Thanks!

PR #62, opened 2020-03-23
Tutorial: Correct ContactInput in generated sdl.

ybakos:
The initial listing of contacts.sdl.js has a typo: it uses 'type'
for ContactInput, when it should be 'input'.

Replace 'type' with 'input' in the code snippet of contacts.sdl.js,
to match the generated gql schema.

See https://github.com/redwoodjs/redwood/issues/305#issuecomment-602308607
See https://github.com/redwoodjs/redwood/issues/317

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -1749,7 +1749,7 @@ export const schema = gql`
     contacts: [Contact]
   }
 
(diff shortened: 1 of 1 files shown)


cannikin:
Thanks!

PR #63, opened 2020-03-23
Tutorial Admin: Preserve generated route order.

ybakos:
The code listing for Routes.js lists the posts routes in
an order that does not match what the scaffold generator
produces. This makes it a little confusing when comparing
the snippet to the existing lines in the actual Routes.js.

Fix the order of the lines to match the contents of Routes.js
in an actual project that follows the tutorial.


cannikin:
Thanks!

PR #66, opened 2020-03-24
Fix html attribute typo

pavelloz:
... when OCD kicks in.


thedavidprice:
ha! OCD… happens to the best of us 😄 

Thanks!

PR #67, opened 2020-03-25
Lazyload hljs

pavelloz:
Lazyloading hljs when there is code tag on a page might make a big difference when it comes to initial page load, especially that:
* hljs is a 3rd party dependency
* its very big
* it doesnt change very often

^ Which means its a great candidate to be cached by the browser.

---

Results in dev mode

Before:
image

After:
image

Those proportions are healthier, IMO :)

PS. I thin netlify preview should use production build to be more suitable for QA, it looks like assets are in dev mode, hence everything appears to be much slower than it will be in reality (hi Tailwind :) ).

### DIFF
--- a/code/javascripts/controllers/application_controller.js
+++ b/code/javascripts/controllers/application_controller.js
@@ -1,5 +1,4 @@
 import { Controller } from "stimulus";
-import hljs from "highlight.js";
 
 export default class extends Controller {
   static get targets() {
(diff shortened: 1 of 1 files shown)


cannikin:
I'm getting a console error the first time I visit a page that had a code block: https://deploy-preview-67--redwoodjs.netlify.com/tutorial/prerequisites

image

pavelloz:
404 is never a good sign, especially if it automagically fixes itself upon refresh :)

cannikin:
I'm in the same boat—I don't know enough about Netlify's build process to debug this. I'll keep it in mind for the future and see what I can do.

PR #68, opened 2020-03-25
[Tutorial] Unnecessary escape in email regexp

olance:
Nothing too big, but `.` does not have to be escaped in character classes


cannikin:
Thanks!

PR #69, opened 2020-03-27
fix typos

Mudassar045:
I was going through `redwoodjs` docs and it's pretty awesome. Just found minor typos.


cannikin:
Thanks!

PR #75, opened 2020-03-30
Add missing `yarn install` step to tutorial

joshsaintjacque:
When going through the tutorial the user is directed to `yarn redwood dev`, which will throw this error:

```
error Command "redwood" not found.
```

In order to avoid this, the user needs to run `yarn install` first. This change updates the tutorial to reflect that.

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -56,9 +56,13 @@ We'll use yarn ([yarn](https://yarnpkg.com/en/docs/install) is a requirement) to
 
     yarn create redwood-app ./redwoodblog
 
-You'll have a new directory `redwoodblog` containing several directories and files. Change to that directory and let's create the database, and then start the development server:
+You'll have a new directory `redwoodblog` containing several directories and files. Change to that directory and install the dependencies:
 
     cd redwoodblog
+    yarn install
+    
+Now, let's create the database, and then start the development server:
+
     yarn redwood dev
 
 A browser should automatically open to http://localhost:8910 and you will see the Redwood welcome page:


joshsaintjacque:
I also notice it mentioned creating the database prior to starting the dev server. Do we want to add a `yarn redwood db up` step before `yarn redwood dev` or remove the reference to creating the DB?

cannikin:
Part of `yarn create redwood-app` is that `yarn install` is run for you in the target directory. You shouldn't need to run it manually at this step (or at all, unless you install your own packages). Did you `cd` into the created directory before running `yarn redwood dev`?

cannikin:
> I also notice it mentioned creating the database prior to starting the dev server. Do we want to add a `yarn redwood db up` step before `yarn redwood dev` or remove the reference to creating the DB?

I think that text is from an old version of the tutorial and never updated...it doesn't actually create the database at that step, but it does create the prisma client libraries for talking TO the database. But the reader doesn't need to worry about that at this point so I'd just remove that text completely and say "start the development server."

joshsaintjacque:
> Part of `yarn create redwood-app` is that `yarn install` is run for you in the target directory. You shouldn't need to run it manually at this step (or at all, unless you install your own packages). Did you `cd` into the created directory before running `yarn redwood dev`?


Thanks for the reply!

Ok, I think I see what happened. I had a previous `yarn create redwood` that failed so the directory was already then during my subsequent attempt. I believe that the install didn't complete because of that, hence why I had to run `yarn install` manually.

Closing...

PR #77, opened 2020-04-02
Update TUTORIAL.md

jrhorn424:
Improve the grammar a bit, IMO. The previous structure was difficult to parse, and I found this more natural. Does it convey the same meaning?


cannikin:
Thanks!

PR #79, opened 2020-04-04
Fix tutorial links

michelegera:
This fixes a couple of links: an example link for the blog tutorial app, and an outdated link to Prisma docs for ID fields.


cannikin:
Thank you much!

PR #85, opened 2020-04-11
Move `useMutation` into the existing import

cball:
I'm unsure if this was intentionally separated, but `@redwoodjs/web` is already being imported for the Form helpers. I think it's best to also export `useMutation` from there since it's what most auto-import plugins would do.

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -1877,8 +1877,8 @@ import {
   Submit,
   FieldError,
   Label,
+  useMutation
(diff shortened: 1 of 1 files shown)


thedavidprice:
Thanks again!

It’s likely done that way to make it explicit to the reader that there’s a change regarding the additional import.

Looping in @cannikin here.

cball:
@cannikin good catch, thanks. I've updated that section as well.

cannikin:
Thanks!

PR #86, opened 2020-04-11
Edit import order to reflect eslint settings

cball:
Since many editors try to auto-fix eslint errors on save, this updates the import order to reflect what a user completing the tutorial might see.


thedavidprice:
Ah, good catch! That import was a recent addition.

PR #87, opened 2020-04-11
Update database section

cball:
This PR tries to add more detail to the database section of the tutorial. I know there's a balance between an easy tutorial and additional setup, but since the concept of Postgres for production was already being introduced, I figured it was worth documenting the local setup required to run that properly as well.

It includes:
* Running Postgres locally (recommended)
* A link to the proper `connection_limit` parameter
* Detail on how to switch between SQLite for local and Postgres for production should the user choose to do so
* More detail on connection pooling and a link for Heroku

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -2216,15 +2216,6 @@ datasource DS {
 Add a `DATABASE_URL` to your `.env` file with the URL of the database you'd like to use locally. We've chosen `redwoodblog_dev` and have `postgres` as a superuser of our local database for ease of use.
 
 ```env
-# THIS FILE SHOULD NOT BE CHECKED INTO YOUR VERSION CONTROL SYSTEM
-#
-# Environment variables set here will override those in .env.defaults.
-# Any environment variables you need in production you will need to setup with
-# your hosting provider. For example in Netlify you can add environment
-# variables in Settings > Build & Deploy > environment
-#
-# DATABASE_URL=postgres://user:pass@postgreshost.com:5432/database_name
-# BINARY_TARGET=rhel-openssl-1.0.x
 DATABASE_URL="postgresql://postgres@localhost/redwoodblog_dev?connection_limit=1"
 ```
 
@@ -2252,16 +2243,6 @@ datasource DS {
 ```
 
 ```env
-# .env
-# THIS FILE SHOULD NOT BE CHECKED INTO YOUR VERSION CONTROL SYSTEM
-#
-# Environment variables set here will override those in .env.defaults.
-# Any environment variables you need in production you will need to setup with
-# your hosting provider. For example in Netlify you can add environment
-# variables in Settings > Build & Deploy > environment
-#
-# DATABASE_URL=postgres://user:pass@postgreshost.com:5432/database_name
-# BINARY_TARGET=rhel-openssl-1.0.x
 DATABASE_PROVIDER="postgres"
 ```
 
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -2246,7 +2246,7 @@ Now just authorize Netlify to connect to your git hosting provider and find your
 
 Netlify will start building your app (click the **Deploying your site** link to watch the logs) and it will say "Site is live", but nothing will work. Why? We haven't told it where to find our database yet.
 
-Go back to the main site page and then to **Settings** at the top, and then **Build & Deploy** > **Environment**. Click **Edit Variables** and this is where we'll paste the database connection URI we got from Heroku (note the **Key** is "DATABASE_URL"):
+Go back to the main site page and then to **Settings** at the top, and then **Build & Deploy** > **Environment**. Click **Edit Variables** and this is where we'll paste the database connection URI we got from Heroku (note the **Key** is "DATABASE_URL"). After pasting the value, append `?connection_limit=1` to the end. The URI will have the following format: `postgres://<user>:<pass>@<url>/<db>?connection_limit=1`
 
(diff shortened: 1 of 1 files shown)


thedavidprice:
@cball this is great and appreciate you taking the time! This information will be _very_ helpful to people right now. 

However, the Tutorial is not the right place for this. But there are two options where this could go before we have a  proper documentation setup including a “How To” section.
1. Add this as a separate markdown file in the RedwoodJS repo [`redwoodjs/redwood/docs`](https://github.com/redwoodjs/redwood/tree/master/docs). Maybe as `localPostgresSetup.md`. Just make it descriptive.
2. In our Forums in the [“Troubleshooting & How To”](https://community.redwoodjs.com/c/troubleshooting/8) section with a descriptive title.

Both places are helpful. But if choosing one, choose “/docs” as it will then be migrated to our improved documentation in the future.

cball:
@thedavidprice thanks! I agree that `/docs` is a better choice than the forum, as it will be closer to the installation instructions that most users will be referencing.

Regarding the tutorial not being the right place for this, I'm assuming you're referring specifically to the local Postgres changes? Otherwise we may want to consider moving the `A Note About DB Connections` section as well.

Assuming we move it, a few questions:
- What should I do with the `connection_limit` parameter documentation? Should that stay in this PR?
- Should I add a one liner in the tutorial to reference the local Postgres setup for clarity? 

The reason I believed it would be good to include these changes is because it was one of the first questions that came to mind as I was going through the tutorial. It was a little confusing that we run SQLite locally, and then Redwood is preconfigured to use the Netlify prisma plugin that makes Postgres work when deployed.

thedavidprice:
All great points and I’ll try to clarify further:
- the Tutorial itself should branch topics infrequently, especially when it comes to other tools/tech that is not specifically necessary to make the Tutorial work. It’s a ton of information and learning. And we need to make it accessible to people early in their dev understanding as well as informative and engaging to all.
- when we offer asides/branches, the are typically of 2 varieties: 1) to explain how Redwood is working under the hood and 2) to provide a bit of a “this isn’t ready for full production” warning

Prisma ships with sqlite. So it’s local dev appropriate and requires no further explanation/setup in the tutorial.

**Linking to the Postrgres Doc from the Tutorial**
This is 100% a great idea. A one sentence, "if you’d prefer to …””  —> Yes, please!

**A Note about DB Connections Section**
It’s just too long. It’s not mean to explain how connection pooling works or how you would setup. It’s meant to be a “heads up!” about potential issues for production requirements.

But _it’s definitely information and helpful for those who want to learn how_! So what about editing the current to be similar length but, perhaps, more to the point. And then a link to “how you’d do this” in the new markdown in /docs.

----

Helpful and making sense?

cball:
@thedavidprice sounds good! Makes perfect sense.

cball:
I've updated to reflect changes we discussed. Note: this PR is currently pointing to a commit hash from redwoodjs/redwood#422 for easier review. Once that PR is merged, I will update this PR to point to master.

TUTORIAL.md
```diff
@@ -2247,6 +2250,8 @@ Go back to the main site page and then to **Settings** at the top, and then **Bu
 
 ![image](https://user-images.githubusercontent.com/300/76902309-f41a5a80-6858-11ea-974f-cbc00863e5a9.png)
 
+> When configuring a production database, you'll want to append `?connection_limit=1 to the URI. This is [recommended by Prisma](https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-client/deployment#recommended-connection-limit) when working with relational databases in a Serverless context.
```
cannikin, line 2253: Can we add a mention of appending the `?connection_limit=1` in the section where we actually copy the ENV var from Heroku and paste it into Netlify? Having the note here is great but there's so much going on it'd be easy to overlook when we get to the part where it actually needs to go.

cball: @cannikin to confirm, you'd like it to live above the "environment variables" screenshot rather than below?

image

cannikin: I like the explanation where it is now, I was just thinking an extra mention of it right there when they're actually pasting in the URL 👆

cball: 👍cool. Will update.

cball: @cannikin updated!

TUTORIAL.md
```diff
@@ -2283,7 +2288,9 @@ Another neat feature of Netlify is _Branch Deploys_. When you create a branch an
 
 ### A Note About DB Connections
 
-In this tutorial, your lambda functions will be connecting directly to the Postgres database. Because Postgres has a limited number of concurrent connections it will accept, this does not scale very well. The proper solution is to put a connection pooling service in front of Postgres and connect to that from your lambda functions. We are working on making this process much easier, but keep it in mind before you deploy a Redwood app to production and announce it to the world.
+In this tutorial, your lambda functions will be connecting directly to the Postgres database. Because Postgres has a limited number of concurrent connections it will accept, this does not scale very well. The proper solution is to put a connection pooling service in front of Postgres and connect to that from your lambda functions. To learn how to do that, see [Connection Pooling](https://github.com/redwoodjs/redwood/blob/ed0eb079ff9f7a67ae7f2ff49d4aef0649a1510e/docs/connectionPooling.md).
```
cannikin, line 2291: Want to just point Connection Pooling to the doc page on the main website? https://redwoodjs.com/docs/connection-pooling

cannikin:
@cball Thanks for the updates! Just left one comment on the Connection Pooling link!

cball:
@cannikin all set. I updated the link to `https://redwoodjs.com/docs/local-postgres-setup` as well.

cannikin:
AWESOME  Thanks so much for this one!!

PR #88, opened 2020-04-13
Add two new DB docs

thedavidprice:
1. local postgres setup
2. settings/config for connection pooling

Did I do this correctly, RC?


thedavidprice:
@cannikin I merged the fix for parser error on redwood side of thing. I’ll let you take this from here.

PR #91, opened 2020-04-19
Add Linux setup instructions

benmccann:
There wasn't much reason to split out by OS for Yarn because their install instructions automatically detect your OS and show you the appropriate instructions. For Node, it was a little more complicated


cannikin:
Thanks!

PR #103, opened 2020-04-23
add Env Var doc

thedavidprice:


cannikin:
> @cannikin Did I do it right?
> 
> [edit: nope... I'll figure it out]

haha Yeah take out the `blob/master` part of the URL.

I'm going to move those docs to the .com repo today so that should make it even simpiler!

PR #105, opened 2020-04-24
improve the news-cards

guledali:


guledali:
@cannikin It would nice if all of the cards had the same heights, is there a way shortening the description?

cannikin:
@guledali Thanks for the PR! I actually wanted the cards different heights, it's a stylistic choice I made. And I don't like zooming in on the logos and cutting them off, I want to see the whole image. I don't think you can enforce a limit to the line length in a description because with a responsive layout you never know where the lines are going to break...some will always end up one line longer or shorter at some browser width.

I actually had a version which tiled them per column instead of per row which looked much better, but I wasn't happy with the chronological order change of the articles when I went from desktop to mobile view. I think I have a solution for that, though.

Thanks again for the PR but I'm going to close this one and apply these changes 👆this morning! 😀

PR #131, opened 2020-05-19
Update postgres setup

chaharshivam:
Related [redwoodjs/redwood#508](https://github.com/redwoodjs/redwood/issues/508)

- Updates the `Local Postgres Setup` Guide with:
  - An installation link for Postgres
  - How to create a user in Postgres
  - Explanation on the connection string
  - Link to database management tools
  -  A note on Postgres manual start/stop

Let me know if you have any suggestions :)


cannikin:
Thank you!

PR #134, opened 2020-05-21
This wording is less than ideal

schacon:
This is one of the more important pull requests you'll probably get on this project.


cannikin:
hahaha Thanks Scott!

PR #140, opened 2020-05-22
Update link to redwood-tools reference

jtoar:
We changed the `redwood-tools` reference section's name in https://github.com/redwoodjs/redwood/pull/577 so updating the link to it here.


cannikin:
Awesome thanks

PR #142, opened 2020-05-23
[Tutorial] Saving Data page - updates line highligting in diff

tmr08c:
Adds additional line highlighting on the "Display Server Errors" section
on the "Saving Data" page of the tutorial.

The section above mentions updating the `Form` component:

> We'll also pass error to \<Form\> so it can setup a context

but the line with the `Form` component was not highlighted. This commit
highlights the line with the `Form` component to make it more obvious
this line needs to be updated as well.

## Before

Screen Shot 2020-05-23 at 5 24 48 PM

## After

Screen Shot 2020-05-23 at 5 36 42 PM


cannikin:
Thank you!

PR #143, opened 2020-05-24
[TUTORIAL] Adds notes about upgrading to Auth section

### LINKED ISSUE #141: Add note about upgrading to 0.6.0 in the Auth section
A lot of users seem to be getting stuck adding Auth because they didn't realize they needed to upgrade to at least 0.7.0. Let's add a little callout a the top of the Authentication section of the tutorial to remind them.

tmr08c:
Closes #141

Adds a call out on the authentication page that `yarn rw g auth`requires Redwood >= 0.7.0.

Screen Shot 2020-05-24 at 12 21 16 PM

## Note on placement

I decided to put it after `yarn rw g auth netlify` because that is where I got confused when running through the tutorial.

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -2410,7 +2410,16 @@ There are a couple of places we need to add some code for authentication and luc
 yarn rw g auth netlify
 ```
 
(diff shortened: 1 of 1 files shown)


cannikin:
Awesome thanks!

PR #146, opened 2020-05-26
Add WIP newsletter

jtoar:
This is an example of a more-legit newsletter. I'm adding it to the website because 1) I used the website's styling 2) if we go ahead with a newsletter like this, maybe we'd add them to the website anyway?

But the upshot of this PR is: there's a new section in the nav (newsletter) with the more-legit newsletter:

newsletter

One for-sure todo is fixing the formatting in "on this page" aside.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -90,6 +90,12 @@ const SECTIONS = [
       },
     ],
   },
+  {
+    name: 'newsletter',
+    files: [
+      { pageBreakAtHeadingDepth: [1], url: './newsletter/wip-newsletter.md' },
+    ],
+  },
 ]
 
 const asyncForEach = async (array, callback) => {
--- a/lib/build.js
+++ b/lib/build.js
@@ -27,50 +27,59 @@ const SECTIONS = [
         url: 'redwoodjs/redwood/README.md',
         title: 'Introduction',
         skipLines: 4,
+        searchable: true,
       },
-      { pageBreakAtHeadingDepth: [1], url: 'redwoodjs/redwood/packages/auth/README.md' },
-      { pageBreakAtHeadingDepth: [1], url: './docs/assets.md' },
+      { pageBreakAtHeadingDepth: [1], url: 'redwoodjs/redwood/packages/auth/README.md', searchable: true, },
+      { pageBreakAtHeadingDepth: [1], url: './docs/assets.md', searchable: true, },
       {
         pageBreakAtHeadingDepth: [1],
         url: './docs/builds.md',
+        searchable: true,
       },
       {
         pageBreakAtHeadingDepth: [1],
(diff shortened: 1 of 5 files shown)


peterp:
This is awesome!

lib/build.js
```diff
@@ -10,7 +10,7 @@ const octokit = Octokit({
 })
 
 const { create: createDocs, buildNav } = require('./docutron.js')
-const { publish: publishSearch, getObjectIDs } = require('./search.js')
+// const { publish: publishSearch, getObjectIDs } = require('./search.js')
```
cannikin, line 13: Make sure this doesn't become part of the real PR! 😬

lib/build.js
```diff
@@ -145,7 +151,7 @@ const setPageVersions = (version) => {
 }
 
 const main = async () => {
-  const objectIDs = await getObjectIDs()
+  // const objectIDs = await getObjectIDs()
```
cannikin, line 154: Likewise

lib/build.js
```diff
@@ -162,7 +168,7 @@ const main = async () => {
       const pages = createDocs(markdown, section.name, { ...file, version })
 
       navLinks += buildNav(pages, section.name, index).join('\n')
-      await publishSearch(markdown, section.name, { ...file, objectIDs })
+      // await publishSearch(markdown, section.name, { ...file, objectIDs })
```
cannikin, line 171: Likewise x2

cannikin:
I'm thinking we'll want to exclude the Newsletter from search, otherwise they'll start dominating the search results. Maybe we add some kind of `searchable` key to the big `SECTIONS` object, which is `true` for everything there now, but `false` for the newsletter section. Then check for that when compiling the sections for search.

lib/build.js
jtoar: Ah--super sorry! (x3)

lib/build.js
```diff
@@ -162,7 +180,7 @@ const main = async () => {
       const pages = createDocs(markdown, section.name, { ...file, version })
 
       navLinks += buildNav(pages, section.name, index).join('\n')
-      await publishSearch(markdown, section.name, { ...file, objectIDs })
+      if (section.searchable) await publishSearch(markdown, section.name, { ...file, objectIDs })
```
jtoar, line 183: @cannikin Is something like this what you meant?

jtoar, line 183: @cannikin Is something like this what you meant? (Sorry I resolved the first one, didn't know what I was doing add-single-comment wise.)

cannikin: Yup! Love it.

jtoar:
Going to close this one; served its purpose on the road to newsletter v0.0.1.

PR #149, opened 2020-05-27
Combine guides and refs to docs

jtoar:
This PR 1) restructure files, combining guides and refs into docs, and (will eventually) 2) update the content of all docs + search.

This first commit's restructures files. All following commits will be about content and search.


docs/authentication.md
```diff
@@ -0,0 +1,238 @@
...
+    </Button>
+  )
+}
+```
+
+## API
+
+The following values are available from the `useAuth` hook:
+
+* async `login()`: Differs based on the client library, with Netlify Identity a pop-up is shown, and with Auth0 the user is redirected
```
cannikin, line 123: It's actually `logIn()` now

docs/authentication.md
```diff
@@ -0,0 +1,238 @@
...
+* async `logout()`: Log out the current user
```
cannikin, line 124: `logOut()`

docs/authentication.md
```diff
@@ -0,0 +1,238 @@
...
+* `currentUser`: an object containing information about the current user, or `null` if the user is not authenticated
+* async `getToken()`: returns a jwt
+* `client`: Access the instance of the client which you passed into `AuthProvider`
+* `authenticated`: used to determine if the current user has authenticated
```
cannikin, line 128: `isAuthenticated()`

docs/authentication.md
jtoar: Thanks! I'll make those changes in the auth readme too.

thedavidprice:
Title Suggestion (for keywords):
"Command Line" --> "CLI Commands"

?

jtoar:
@thedavidprice Where can I figure out how keywords work exactly? Just the algolia docs?

thedavidprice:
I'm actually not sure how Algolia settings work. That would be @cannikin's domain. So far it seems to index on page titles and/or menu items.

jtoar:
@cannikin This one's ready for review. I've organized things so that this PR only changes the file structure, combining Guides and the Reference to Docs, + some changes to the website where appropriate. E.g., the "Read the Guides" button now says "Read the Docs", but still points to the Introduction (now in Docs).

After this gets merged, I'll make a tracking issue for the Docs to organize the effort / launch several more PRs.

Let me know what you think about this modularity, if I should include any other changes here / take some out. Thanks!

package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.1",
   "scripts": {
     "build": "node lib/build.js && webpack && postcss --verbose code/stylesheets/application.pcss -o publish/stylesheets/application.css",
-    "clean": "del 'code/html/(cookbook|guides|tutorial|reference)/*.html' 'code/html/_(cookbook|guides|tutorial|reference)_nav.html' 'publish/!(downloads|images|favicon.*)'",
+    "clean": "del 'code/html/(cookbook|docs|tutorial|reference)/*.html' 'code/html/_(cookbook|docs|tutorial|reference)_nav.html' 'publish/!(downloads|images|favicon.*)'",
```
cannikin, line 6: Can you remove "reference" from here now?

netlify.toml
```diff
@@ -27,15 +27,15 @@
   environment = { NODE_ENV = "production" }
 
 [[redirects]]
-  from = "/guides"
-  to = "/guides/introduction"
+  from = "/docs"
```
cannikin, line 30: Can you remove any "reference" redirects/headers?

cannikin:
If you check the [preview deploy](https://deploy-preview-149--redwoodjs.netlify.app/) there are two "Command Line Interface" links and clicking the second one just brings you to the first one. I assume that's because of the two separate docs, but now that the titles are the same the URLs are the same as well. We could just remove one for now, so we can get this merged and live, then combine the content of the removed one into the other in another PR.

Screen Shot 2020-05-29 at 12 50 32 PM

package.json
jtoar: Yes, totally missed that thanks.

netlify.toml
jtoar: Removed!

cannikin:
Looks great, thanks for doing this @jtoar!

jtoar:
No problem! 

@cannikin One more thing though--I just noticed that the search still lists results as being "Guides", and the links are broken:

rw-search

Could you point me in the right direction to help on this? Did I miss something in `search.js`? Looking now.

cannikin:
I'll have to wipe out the index and re-deploy to rebuild it. I'm pretty good about cleaning up search items that don't exist any more, but if we remove an entire section like "Guides" my code doesn't cover that. :/

cannikin:
Try now!

jtoar:
> I'm pretty good about cleaning up search items that don't exist any more, but if we remove an entire section like "Guides" my code doesn't cover that. :/

Doesn't happen too often! :sweat_smile:

PR #155, opened 2020-05-29
Copy over and rename/reorder files

jtoar:
This PR 1) copies over any files that were referenced from the repo and 2) re-names/orders all the files so that they're a) titled after their heading and b) in alphabetical order.

@cannikin Wanted to do this before opening a giant tracking issue / recruiting others. Things will probably get renamed (and then reordered) again after we start work on them and esp. after we target searchability. But at least things will be in order till then!

PR #157, opened 2020-05-30
add reference + guides --> docs redirects per #149

thedavidprice:
re: https://community.redwoodjs.com/t/the-guides-are-gone/652

@cannikin not sure if this is the cleanest way to do it or not, but just trying to get things working for now

@jtoar this one's on me, sorry. Web 101 -- if you move things, made sure the Internet can still find 'em! Let make sure we keep these redirects updated as things move forward with the restructuring.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -85,7 +85,7 @@ to = "/"
 [[redirects]]
   from = "/guides/assets-and-files"
   to = "/docs/assets-and-files"
-  
+
 [[redirects]]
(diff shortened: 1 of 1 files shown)


thedavidprice:
tested 'em all and working 🚀

jtoar:
@thedavidprice Ah super sorry! And right--don't break the web. I'll take another look at #155 because there's some title changes there. Thanks for catching this!

PR #161, opened 2020-05-31
Demo tables

jtoar:
> Note: not meant to be merged / forever a draft

https://deploy-preview-161--redwoodjs.netlify.app/docs/tables

This PR demos some table designs. When I started writing docs, I seemed to be adding tables left and right. So I figured we'd better try to agree on a design up front. But I'm sure we'll still tweak the design slightly as we actually add them to the docs. 

I think [design 3](https://deploy-preview-161--redwoodjs.netlify.app/docs/tables#3-background-header) was a favorite on the last call but @cannikin @thedavidprice @mojombo @peterp what do you guys think?

### DIFF
--- a/code/stylesheets/markdown.pcss
+++ b/code/stylesheets/markdown.pcss
@@ -1,11 +1,6 @@
 /* I'm used on the built-in CameronJS welcome page code/html/index.html */
 /* Feel free to delete me once you start changing that page.            */
 
-#code {
-  @apply bg-red-200 text-gray-900 text-sm px-1 py-half rounded;
-	font-family: Fira Code, Fira Mono, Menlo, Monoco, monospace;
-}
-
 #link {
   @apply text-red-700
 }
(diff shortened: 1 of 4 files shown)


mojombo:
My favorite is 3a. It visually keeps things separated, and doesn’t use the link color for plain text. I also like the normal size text. Curious what @cannikin thinks. Thanks for putting these options together!

cannikin:
3c for me, I like the code being colored. Even though the the color matches the link color, the background also being colored made it seem separate enough from a link for me. Either way, maybe we should underline links. I’m surprised I didn’t do that to start with!

jtoar:
We do use link-color-for-plain-text + code formatting on the home page:

Frame 1

mojombo:
That’s a fair point. I do like 3c. Let’s go with that!

thedavidprice:
Great work on this, Dom! One thing to watch out for is the left column size when viewed on mobile responsive. It's line-breaking the "--" from "verbose" in this case. A few possible fixes, including responsive text size and css `white-space: nowrap;`. But probably best to primarily increase the column width itself.

More thoughts to this end, RC?

Screen Shot 2020-05-31 at 11 33 22 AM

cannikin:
Maybe we can add something like 

```css
td code { 
  @apply whitespace-no-wrap 
}
```

so that it only no-wraps them inside the tables, otherwise the code will get no-wrapped everywhere, which would probably be bad.

jtoar:
@thedavidprice Do you still see the line-breaking with the tables on https://deploy-preview-161--redwoodjs.netlify.app/docs/tables-continued?

code/stylesheets/markdown.pcss
```diff
@@ -105,4 +100,52 @@
   h6:hover .markdownIt-Anchor {
     visibility: visible;
   }
+
+  table {
```
jtoar, line 104: @cannikin Starting here till line 127, I added markdown table-specific css. Does this look right? (Ignore the `#code` etc towards the bottom. Those were just for demoing multiple designs.)

cannikin: Looks good to me!

thedavidprice:
This is looking really good to me (shown for iPhone 6/7/8 screen emulator):
Screen Shot 2020-06-01 at 10 01 31 AM

@jtoar You might already know about this, but Google Developer Tools has a "toggle device toolbar" option. Icon to enable below or `shift + cmd + M`
Screen Shot 2020-06-01 at 10 02 53 AM

jtoar:
@thedavidprice Thanks! I did try it out but I don't trust that I actually know how to use it. :sweat_smile:

jtoar:
Safe to close this one?

mojombo:
@jtoar yep, let's move forward with these styles. Thanks!!

cannikin:
Yup I think so!

Issue #156, opened 2020-05-30
Improvements to Existing Docs (tracking issue)


jtoar:
# Improvements to Existing Docs (tracking issue)

The following's a list of all the docs we got, what they're about or should be about, and questions/helpful information. I'll continue to update it as things progress:

- [ ] **App Configuration: redwood.toml:** Should list all the configuration options and detail certain setups (e.g. container/vm). Should discuss the importance of the file in identifying a redwood project as a redwood project. The concept of sides/targets is intrinsic to this file, so there's some room for explanation / discussion about the future.
  - This is only for the dev environment right?
  - Are there other ways to configure your app besides toml file? If so then this file should talk about them? Or this file's just about redwood.toml? Just about configuring the dev environment? There's netlify.toml, but we won't talk about that here?
- [ ] **Assets and Files:** Should detail the two ways to add assets to a Redwood app, pros/cons of each, or moreover, when you'd want to use each one. Could use more examples.
- [ ] **Authentication:** This is one of those docs will only get bigger and bigger. We might want to look into adding mdx-like "tabs" at some point? I.e., tabs that let you switch between providers--to avoid having an infinitely long doc.
- [ ] **Builds:** Should...?
- [ ] **CLI Commands:** All reference right now; should feature more of a usage section--e.g., CLI basics. Need to go into more detail on generators, especially the sdl--where we plan on drawing the line in schema.prisma <-> sdl support...
- [ ] **Connection Pooling:** Combine with Local Postgres Setup into "Database" doc?
- [ ] **Contributing:** Should help readers make the context switch, from developing to contributing by linking to the package-agnostic contributing readme and each package's readme in turn.
- [ ] **Environment Variables:** Should...?
- [ ] **Form:** A model of we're striving for. This one's extremely detailed.
- [ ] **Local Postgres Setup:** Combine into with Connection Pooling into "Database" doc?
- [ ] **Router:** Another great doc/model. Should probably remove sections about installation/use outside redwood apps?
- [ ] **Serverless Functions:** Should...? Also see the Custom Function cookbook...
- [ ] **Webpack Configuration:** This one's a little more outright how-to than reference--how to configure webpack. We could do some explanation though; it might be helpful to know what not to override, etc. There's already some examples here.

## Taking on a Doc 

> Please read the working guidelines: https://github.com/redwoodjs/redwood/issues/332

All the docs on redwoodjs.com have a context: they're for developing. They're here to help people develop apps using Redwood. 

[...]

PR #163, opened 2020-05-31
Indent "On this page" nav

jtoar:
This PR adds indenting to the "On this page" nav. We're talking about this feature in #156. This is a way to see what it might look like/add it if we decide to go ahead. 

_Before_:

rw-indent-before

_After_:

rw-indent-after

### DIFF
--- a/lib/docutron.js
+++ b/lib/docutron.js
@@ -43,7 +43,8 @@ const convertToHtml = (pages, book, version) => {
     const content = md.render(page.text)
     const output = TEMPLATES.page({
       content,
-      links: subNav(page.text),
+      // Remove any top-level headers (i.e. #) b/c they're redundant
+      links: subNav(page.text).filter((header) => header.level > 1),
       nextPage: pages[index + 1],
       pageTitle: `${titleCase(book)} - ${page.title}`,
       version: version,
@@ -117,6 +118,9 @@ const subNav = (markdown) => {
       href: `#${paramCase(title.toLowerCase().replace(/&.*?;/g, ''))}`,
       level: headerLevelIndex,
       title: title,
+      // To indent headers according to the number of # they have.
+      // But since we remove #, which equal 1, we want to consider ## as 1 and indent from there.
+      ml: (headerLevelIndex - 2) * .75
     }
   })
 }
--- a/lib/templates/page.html.template
+++ b/lib/templates/page.html.template
@@ -4,10 +4,12 @@
   <aside class="hidden xl:block w-1/4 h-auto overflow-y-visible">
     <div class="overflow-y-auto sticky top-0 pl-6">
       <nav class="text-sm font-semibold text-gray-600 pt-8 pb-40 lg:py-12 overflow-y-auto h-screen" data-controller="aside" data-action="scroll->aside#saveScroll">
-        <h4 class="text-xs uppercase text-gray-500">On this page</h4>
+        <% if (links.length) {  %>
+          <h4 class="text-xs uppercase text-gray-500">On this page</h4>
+        <% } %>
         <ul class="mt-4 pointer-events-auto">
           <% links.forEach(link => { %>
-          <li class="my-1 py-1">
+          <li class="my-1 py-1" style="margin-left: ${link.ml}rem;">
             <a href="${link.href}">${link.title}</a>
           </li>
           <% }) %>


thedavidprice:
I'm a big fan of this indentation. 👍
- are the items below `[api]` not indented due to having same headers for `api.host`, etc?
- including the h1 in the menu is redundant, e.g. we should remove the `App Configuration: redwood.toml`. I recognize there is often introductory content located within this section but 1) when you first visit the page you're already there and 2) if you entered the page via an anchor link, it’s not like the h1 is actually telling you what the intro content is, specifically, it would just be a way to get to the top of the page.

jtoar:
1. Yes, the doc looks like this:
   
```markdown
## [api]

This table contains the configuration for api side.

## api.host

The hostname (string) to listen to for the web server, defaults to `localhost`. When running the server within containers/VMs, using `0.0.0.0` would allow network connections to/from the host.

## api.port

The port number (integer) to listen to for the api side.
```

2. I just pushed this change up. The nice thing's about it is the tutorial stays unaffected since sections are `##`. The [Welcome to Redwood](https://deploy-preview-163--redwoodjs.netlify.app/tutorial/welcome-to-redwood) now has a blank "On this page" nav, but there's not much there to navigate to anyway.

@thedavidprice would you rather remove the header from tutorial pages too? I think I'm in favor of keeping them since a lot of the content is under the first header there.

thedavidprice:
> would you rather remove the header from tutorial pages too? I think I'm in favor of keeping them since a lot of the content is under the first header there.

Great point and for the Tutorial I 100% agree --> each of the sections are more like "steps" and have additive meaning. We wouldn't want to remove them.

If it's non-trivial to remove from the docs sidebar only, then let's not consider it. But for the docs I think it just adds redundancy and extra indentation levels that aren't really necessary. (Small details/point overall, to be clear.)

jtoar:
@thedavidprice After that last commit we remove `#`'s, and since tutorial pages start with `##` they're unaffected. So I think that box is checked. The other considerations are: 

- Indenting too much/too little?
- Should we remove "On this page" entirely if there's no `#`? Like on [Welcome](https://deploy-preview-163--redwoodjs.netlify.app/tutorial/welcome-to-redwood) (note that this is the only case)

thedavidprice:
Nice!

**Indenting**
For my taste, I think you could bump from .5rem to .75rem. But I have no hard opinion on this.

**Remove "On this page"**
That would be ideal if not too much trouble. Again, small detail (although small details are often the ones that make a difference!).

jtoar:
@thedavidprice "On this page" is now removed for pages w/o links; see [Welcome](https://deploy-preview-163--redwoodjs.netlify.app/tutorial/welcome-to-redwood) for an example. And here's a side by side of 0.5 vs 0.75:

**0.5**

rw-5

**0.75**

rw-75

Still too hard to tell/decide? @cannikin would appreciate your take on this as well!

cannikin:
I like the 0.75 indent!

Do these changes take into account that you can set the `pageBreakAtHeadingDepth` option as part of the build options? It lets you split a single doc into multiple pages, although right now the only doc that uses something other than the default of `[1]` is the tutorial. If we ever decide to use that more extensively, is ignoring the top level title going to mess up other pages?

jtoar:
Not really! We pretty much just remove `#` and start indenting from `##`, but if pages start with `###`, the indenting would probably look weird. I'll convert this to a draft till I figure that out.

cannikin:
We don't actually have anything that wants page breaks at level 3 so I wouldn't put any more effort into it, we can cross that bridge when we come to it! I just wanted to make sure that if some other sections suddenly had a `[1, 2]` that if they behaved like the tutorial does (still includes the ## heading in the nav) we're okay with that!

jtoar:
Ah got it! Id' say yes since if they did they'd most likely contain a lot of content under their second heading, like the tutorial. @thedavidprice What do you think?

thedavidprice:
Agreed, and this looks ready to merge to me (@cannikin I'll let you do so if agreed).

@jtoar When this initiate is complete, we'll likely need YetAnotherDoc™ with information about how to manage, structure, and contribute documentation. It would be a final version of [things included here](https://github.com/redwoodjs/redwood/issues/332). These little decisions and learnings, e.g. right menu indentation picks up `##` and `###`, are the kinds of things that eventually should be included. _Don't_ feel like you need to create that now. But, if you haven't, I'd recommend keeping and adding to a bullet list of things like this to eventually include.

cannikin:
Merged! Thanks for this one, indenting that nav feels very nice!

PR #166, opened 2020-06-01
Decrease "On this page" breakpoint

jtoar:
This PR decreases the "On this page" breakpoint from `xl` to `lg`. Originated from discussion on #156.

### DIFF
--- a/lib/templates/page.html.template
+++ b/lib/templates/page.html.template
@@ -1,7 +1,7 @@
 @@layout("application", { "title": "${pageTitle} : RedwoodJS Docs", "version": "${version}" })
 
 @@contentFor("aside",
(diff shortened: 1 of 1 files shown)


cannikin:
Yeah you're going to have a play with the main content well's width quite a bit. Narrowing the browser down past the `xl` breakpoint just a couple hundred pixels squishes that nav to a ridiculous degree. But we don't want to the readability of the main content to suffer (becoming a column that's only 300px wide, for example) just to include that right nav...

cannikin:
How are you feeling about this PR @jtoar? It looks like the "On this page" nav is much more prevalent on the new learn.redwoodjs.com site with Docusaurus: https://learn-redwood.netlify.app/docs/tutorial/getting-dynamic Eventually all docs will be there so I'd vote to just close this.

PR #167, opened 2020-06-01
Persist scroll position of left nav when changing pages

cannikin:
Re: https://github.com/redwoodjs/redwoodjs.com/issues/156#issuecomment-637077517

This feels much nicer, the nav doesn't jump around as you click on links and sections that you opened/closed stay that way forever (well, until a hard page reload).

/cc @jtoar @thedavidprice @weaversam8

### DIFF
--- a/code/html/layouts/application.html
+++ b/code/html/layouts/application.html
@@ -26,7 +26,7 @@
     <script src="https://cdn.jsdelivr.net/npm/clipboard@2/dist/clipboard.min.js"></script>
   </head>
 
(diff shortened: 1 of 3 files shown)


jtoar:
This feels great! :+1:

cannikin:
Hmm, it works perfectly when the top nav isn't also scrolled off the page, but if it is you get like a 90px jump (the height of the header) when the page loads. Lemmie see if can compensate for that as well...

cannikin:
Ahhh, better.

jtoar:
@cannikin Seriously feels so much smoother. Thanks for this!

weaversam8:
This looks great!

PR #173, opened 2020-06-06
Improve App Configuration doc

jtoar:
> Link to the new-and-improved App Configuration doc: https://deploy-preview-173--redwoodjs.netlify.app/docs/app-configuration-redwood-toml.
> See #156 for the tracking issue for all docs.

This PR improves the content in the [App Configuration](https://redwoodjs.com/docs/app-configuration-redwood-toml) (linking to the original here) doc. I think what I have is a good first draft, but I still need to ask some questions before I can consider it final (sorry this is so long):

### 1. `redwood.toml` is about...

The original doc says:

> Development environment settings can be adjusted using redwood.toml.

Is `redwood.toml` only about configuring the development environment? Would it be more instructive to say it's about configuring _redwood_ settings? Or is it really only about the development environment (like `browser.open` is pretty specific to dev)?

### 2. All the options

Did I get it right that what's in [config.ts](https://github.com/redwoodjs/redwood/blob/d51ade08118c17459cebcdb496197ea52485364a/packages/internal/src/config.ts#L42-L60) is all the options? That is:

```javascript
const DEFAULT_CONFIG: Config = {
  web: {
    host: 'localhost',
    port: 8910,
    path: './web',
    target: TargetEnum.BROWSER,
    apiProxyPath: '/.netlify/functions',
    apiProxyPort: 8911,
  },
  api: {
    host: 'localhost',
    port: 8911,
    path: './api',
    target: TargetEnum.NODE,
  },
  browser: {
    open: true,
  },
}
```

**Update**: `includeEnvironmentVariables` is defined in webpack.common.js, so it's more nuanced than I thought.

### 3. Where do we specify the lambda functions dir? 

In the [Serverless Functions](https://redwoodjs.com/docs/serverless-functions) doc, we say:

> The dev server looks for "lambda functions" in the directory (default: ./api/src/functions) specified by your redwood.toml configuration file.

Where do we specify the lambda functions directory? Maybe [this issue](https://github.com/redwoodjs/redwood/issues/322#issuecomment-634418306) has the answer? Relevant excerpt here:

> 5. "Redwood's HTTP server for serverless Functions" is a mess. (Maybe the title specifically.) Two things:
>   - Move appropriate content to the redwood.toml doc section web.apiProxyPaths

### 4. Why do we only expose these by default?

Why do we only expose these by default in a Redwood app's `redwood.toml`?

```toml
[web]
  port = 8910
  apiProxyPath = "/.netlify/functions"
[api]
  port = 8911
[browser]
  open = true
```

I think the answer is "because you won't be changing the other ones", but just wanted to make sure (for the docs).

Also, I think it'd be nice if `redwood.toml` came commented, like

```toml
[web]
  # Port to listen on
  port = 8910
  # Proxy path to the api side. Our default is the same as netlifys
  apiProxyPath = "/.netlify/functions"
[api]
  # Port listen on...
  port = 8911
[browser]
  # whether or not the browser opens on yarn rw dev
  open = true
```

Maybe this looks a little busy. We don't have to comment on everything; but at the very least we could provide a header of sorts, like .[env.defaults](https://github.com/redwoodjs/create-redwood-app/blob/master/.env.defaults):

```toml
# This file contains the configuration settings for your Redwood app.
# By default, proxy path to the api side is the same as Netlify's
# For the full list of options, see config.ts: https://github.com/redwoodjs/redwood/blob/d51ade08118c17459cebcdb496197ea52485364a/packages/internal/src/config.ts#L42-L60
```

### 5. `browser.open` default

[config.ts](https://github.com/redwoodjs/redwood/blob/d51ade08118c17459cebcdb496197ea52485364a/packages/internal/src/config.ts#L42-L60) says `browser.open` defaults to true, but wanted to double check because I saw this issue come up recently on the forum: https://community.redwoodjs.com/t/stop-yarn-rw-dev-from-opening-browser-window/700/2.

**Update**: saw this line in [webpack.development.js](https://github.com/redwoodjs/redwood/blob/7d1840ab5e5fd02284b31f7077927a9a9b339e54/packages/core/config/webpack.development.js#L31-L32):

```javascript
// checks for override in redwood.toml, defaults to true
open: redwoodConfig.browser ? redwoodConfig.browser.open : false,
```

The comment seems to contradict the code?

### Thanks!

### DIFF
--- a/docs/appConfiguration.md
+++ b/docs/appConfiguration.md
@@ -1,47 +1,143 @@
 # App Configuration: redwood.toml
 
-Development environment settings can be adjusted using `redwood.toml`.
+You can configure your Redwood app's settings in `redwood.toml`. By default, `redwood.toml` lists the following configuration options:
+
+<!-- TODO -->
+<!-- toml syntax coloring not working here -->
+```toml
+[web]
+  port = 8910
+  apiProxyPath = "/.netlify/functions"
+[api]
+  port = 8911
+[browser]
+  open = true
+```
+
+These are listed by default because they're the ones that you're most likely to configure. But there are plenty more available. The rest are spread between Redwood's [webpack configuration files](https://github.com/redwoodjs/redwood/tree/main/packages/core/config) and `@redwoodjs/internal`'s [config.ts](https://github.com/redwoodjs/redwood/blob/main/packages/internal/src/config.ts#L42-L60):
+
+```javascript
+// redwood/packages/internal/src/config.ts
+
+const DEFAULT_CONFIG: Config = {
+  web: {
+    host: 'localhost',
+    port: 8910,
+    path: './web',
+    target: TargetEnum.BROWSER,
+    apiProxyPath: '/.netlify/functions',
+  },
+  api: {
+    host: 'localhost',
+    port: 8911,
+    path: './api',
+    target: TargetEnum.NODE,
+  },
+  browser: {
+    open: true,
+  },
+}
+```
+
+The options and their structure are based on Redwood's notion of sides and targets. Right now, Redwood has two fixed sides, api and web, that target NodeJS Lambdas and Browsers respectively. In the future, we'll add support for more sides and targets, like Electron and React Native (you can already see them listed as enums in [TargetEnum](https://github.com/redwoodjs/redwood/blob/d51ade08118c17459cebcdb496197ea52485364a/packages/internal/src/config.ts#L11-L12)), and as we do, you'll see them reflected in `redwood.toml`. But right now, you'll most likely never touch options like `target`.
+
+The idea is that, in the future, changes here will have cascading, "app-level" effects. Using generators as an example, based on your side and target, the generators will behave differently, but appropriately different.
+
+> For the difference between a side and a target, see [Redwood File Structure](https://redwoodjs.com/tutorial/redwood-file-structure).
+
+You can think of `redwood.toml` as a convenience layer over Redwood's webpack configuration files. That is, for certain settings, instead of having to deal with webpack directly, we give you quick access via `redwood.toml`. Some of these settings are for development, some are for production, and some are for both. You can actually see this reflected in which webpack file each configuraiton option is referenced in&mdash;[webpack.development.js](https://github.com/redwoodjs/redwood/blob/main/packages/core/config/webpack.development.js), [webpack.production.js](https://github.com/redwoodjs/redwood/blob/main/packages/core/config/webpack.production.js), and [webpack.common.js](https://github.com/redwoodjs/redwood/blob/main/packages/core/config/webpack.common.js).
+
+<!-- https://github.com/redwoodjs/redwood/pull/152#issuecomment-593835518 -->
+`redwood.toml` also serves a slightly larger purpose: it's used to determine the base directory of a Redwood project. So this file is what really makes a Redwood app a Redwood app. If you remove it and run `yarn rw dev`, you'll get an error:
+
+```terminal
+Error: Could not find a "redwood.toml" file, are you sure you're in a Redwood project?
+```
+
+(So don't do that!)
 
 ## [web]
 
-This table contains the configuration for web side.
+Configuration for the web side.
 
-### web.host
+| Key                           | Description                        | Default                 | Context       |
+| :---------------------------- | :--------------------------------- | :---------------------- | :------------ |
+| `host`                        | Hostname to listen on              | `'localhost'`           | `development` |
+| `port`                        | Port to listen on                  | `8910`                  | `development` |
+| `path`                        | Path to the web side               | `'./web'`               | `both`        |
+| `target`                      | Target for the web side            | `TargetEnum.BROWSER`    | `both`        |
+| `apiProxyPath`                | Proxy path to the api side         | `'/.netlify/functions'` | `production`  |
+| `includeEnvironmentVariables` | Environment variables to whitelist |                         | `both`        |
 
-The hostname (string) to listen to for the web server, defaults to `localhost`. When running the server within containers/VMs, using `0.0.0.0` would allow network connections to/from the host.
+### apiProxyPath
 
-### web.port
+```toml
+[web]
+  apiProxyPath = "/.netlify/functions"
+```
 
-The port number (integer) to listen to for the web side.
+The path to the serverless functions. When you're running your app locally, this gets aliased away (you can see exactly how in [webpack.common.js](https://github.com/redwoodjs/redwood/blob/49c3afecc210709641dd340b974c86251ed207dc/packages/core/config/webpack.development.js#L21-L28) (and here's the docs on Webpack's [devServer.proxy](https://webpack.js.org/configuration/dev-server/#devserverproxy), for good measure)).
 
-### web.apiProxyPaths
-
-TODO
+Since Redwood plays nicely with Netlify, we use the [same proxy path](https://docs.netlify.com/functions/build-with-javascript) by default. If you're deploying elsewhere, you'll want to change this.
 
 ### includeEnvironmentVariables
+<!-- https://github.com/redwoodjs/redwood/issues/427 -->
+<!-- https://github.com/redwoodjs/redwood/blob/d51ade08118c17459cebcdb496197ea52485364a/packages/core/config/webpack.common.js#L17-L31 -->
 
-The set of environment variable keys (list of strings) to include for the web side, in addition to any that are prefixed with `REDWOOD_ENV_`.
+```toml
+[web]
+  includeEnvironmentVariables = ['API_KEY']
+```
+
+Where `API_KEY` is defined in .env or .env.defaults:
+
+```plaintext
+API_KEY=...
+```
+
+`includeEnvironmentVariables` is the set of environment variables to whitelist for the web side. You can also prefix environment variables with `REDWOOD_ENV_` (see [Environment Variables](https://redwoodjs.com/docs/environment-variables#web)).
 
 ## [api]
 
-This table contains the configuration for api side.
+Configuration for the api side.
 
-## api.host
+| Key      | Description             | Default           | Context       |
+| :------- | :---------------------- | :---------------- | :------------ |
+| `host`   | Hostname to listen on   | `'localhost'`     | `development` |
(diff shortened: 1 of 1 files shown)


thedavidprice:
There's a lot going on in this one. Thanks for taking the first pass! And I expect we'll need a few rounds here -- no problem at all.

## 1. redwood.toml is about...
Correct, this has evolved and will continue to evolve. It's about Redwood settings in general and applies to dev (open browser), application core config (api path, sides), cli (new auto-update check), etc.

## 2. All the options
You got 'em all. And, for the record, I would have forgotten the Env Vars option. That might be worth adding a comment to `config.ts`. Not critical but would be helpful I think.

## 3. Where do we specify the lambda functions dir?
You have it correct; this is the `apiProxyPath`. But, to be clear, this isn't about the location of the functions directory --> this becomes the URL path for the Web side to connect to the API functions. Does that make sense? It's effectively the endpoint for calls to GraphQL server.

This is handled by Webpack at build time. And during local dev it's effectively aliased away with a Webpack path-rewrite using and api `port`. I can show you where/how this works, but easy to search for apiProxyPath and the Webpack files will show up.

Make this section a little stronger --> it's the _required_ path for your API endpoint. The default is set to Netlify, correct.

## 4. Why do we only expose these by default?
I believe it's just to keep things simpler for now (probably mixed with some history). But also some of those hidden defaults will become more important as Sides become extensible. Most people will _not_ be changing the other ones. But, e.g., localhost needs changing if you're using a Virtual sandbox for development. A few others:

**target:** this default resolves to `browser` for web and `node` for API. (They're TS definitions at the top of the file.) These will become more important and make more sense with added Sides, e.g. React Native and Electron...

**port:** for API this is the local dev port, effectively what is aliased to for local dev apiProxyPath

**apiProxyPort:** for Web -- I actually don't believe this is used anymore. Maybe add a question for Peter during the final review.

> # This file contains the configuration settings for your Redwood app.
# By default, proxy path to the api side is the same as Netlify's
# For the full list of options, see config.ts: https://github.com/redwoodjs/redwood/blob/d51ade08118c17459cebcdb496197ea52485364a/packages/internal/src/config.ts#L42-L60

^^ YES, please! Great idea to add this to the top of the file in CRA repo. But let's link to this document instead of the file. 🚀

## 5. browser.open default
Ah, you're correct --> the default behavior is `false`. So if no config then it won't open.

Here's the doc for the Webpack setting --> there's actually quite a bit more people can do if you want to just mention briefly and link to it: https://webpack.js.org/configuration/dev-server/#devserveropen

## Misc
### redwood.toml is required
> Redwood uses redwood.toml to determine the base directory of a Redwood project. If you remove this file and try to run yarn rw dev, you'll get an error:

^^ this is super important if you'd like to visually highlight more

### [web] and [api]
if applicable, you could mention whether an option is for development. E.g. "The development port to listen on"... or something like that. Local dev, dev-server, etc...

### Running within a Container or VM
This gets a little lost here, but I don't know where else to put it and it's not enough for it's own doc. Maybe the tile or section should be "Example Configurations" with setting port for local dev within a Container or VM. 

Right now I can't think of another example. But once we have more deploy targets we'll need to update the apiProxyPath. 

### Accessing Configuration Options from the Web Side
Hmm, we're going to need to loop in Peter on this one as well. For starters, handling Globals just changes yesterday as of this PR https://github.com/redwoodjs/redwood/pull/697 Which might be a breaking change for some people.

But I'm not sure this applies to _all_ globals and _all_ config. I think just apiProxyPath. Also, trying to think of a case where apps might need the other config values in production past build.

thedavidprice:
@jtoar when you're working on Docs next, please prioritize this one -- config Docs + questions have been coming up more frequently amongst community.

jtoar:
@thedavidprice **Re browser.open default:** If we can run over this one more time, what's the reason behind the if statement here in [webpack.development.js](https://github.com/redwoodjs/redwood/blob/7d1840ab5e5fd02284b31f7077927a9a9b339e54/packages/core/config/webpack.development.js#L31-L32)? Because I think it might be superfluous. Reprinting here:

```javascript
// checks for override in redwood.toml, defaults to true
open: redwoodConfig.browser ? redwoodConfig.browser.open : false,
```

My question is: won't `redwoodConfig.browser.open` always be defined, making the check for it above unnecessary? Because in config.ts, we [always do a deep merge](https://github.com/redwoodjs/redwood/blob/d51ade08118c17459cebcdb496197ea52485364a/packages/internal/src/config.ts#L69
) of the Redwood app's redwood.toml with DEFAULT_CONFIG:

```javascript
return merge(DEFAULT_CONFIG, toml.parse(rawConfig))
```

and DEFAULT_CONFIG always defines browser.open. 

So basically what I'm saying is the browser will always open unless you explicitly set it to false. Here's what I tried:

1. Remove just open from [browser] in redwood.toml: browser opens
2. Remove the entire [browser] table in redwood.toml: browser opens
3. Remove all the contents in redwood.toml but not the file itself: browser opens
4. Set open to false in redwood.toml: browser doesn't open

thedavidprice:
Ah, I see your confusion and what's going on --> if you view the history (blame) for the `return merge(DEFAULT_CONFIG, toml.parse(rawConfig))` you'll see that was added recently, at which time the default behavior **should have been set to false** (as originally designed). 

**Expected Behavior**
OPEN
- only in case of `[browser] open = true`

NO EFFECT
- `[browser] open = false`
- no config give in `redwood.toml`

---

Does that help clear things up? And, would you like to update config.ts and the webpack.development or should I? (I'm happy to do so.)

jtoar:
Yes all clear here now. Thank you! And don't worry I'll update them. :+1:

peterp:
> is proxy more of a dev thing than a production thing?

It's for both - although the complexity of that might be extracted away from you on Netlify. The static content is hosted in one place, and they provide a proxy via `/.netlify/functions/<funcName>` to the lambda functions.

The reason why we have it as a configuration option is because on vercel it's `/api/`, which I personally think is very nice and would like to make the default once we're supporting more deployment platforms.

But, I'm open to changing the name, or if we decide to use `/api/` then hiding the option for user's completely.

It would be nice to support a full hostname: `apiProxyPath = https://api.mysite.com/<funcName>`, and then "path" doesn't make sense, maybe `apiUrl`?

jtoar:
> It's for both - although the complexity of that might be extracted away from you on Netlify. The static content is hosted in one place, and they provide a proxy via /.netlify/functions/<funcName> to the lambda functions.

That makes sense. And I don't mind the "path" in "apiProxyPath" as we use "path" in the router to mean basically the same thing. Thanks for clearing up @peterp!

thedavidprice:
@jtoar This one's looking good! Thanks for keeping it moving forward.

jtoar:
@peterp @thedavidprice This one's ready to merge pending approval. The one thing I'm not crazy about is the `web` table. Adding the "Context" column (to tell readers if the variables used in development, production, or both) is squishing the "Description" column.

Should I just make the table scroll? Or any other ideas?

docs/appConfiguration.md
```diff
@@ -1,47 +1,143 @@
...
-The hostname (string) to listen to for the web server, defaults to `localhost`. When running the server within containers/VMs, using `0.0.0.0` would allow network connections to/from the host.
+## [browser]
 
-## api.port
+Configuration for the browser target.
 
-The port number (integer) to listen to for the api side.
+| Key    | Description                                  | Default | Context       |
+| :----- | :------------------------------------------- | :------ | :------------ |
+| `open` | Open the browser after the dev server starts | `false` | `development` |
```
thedavidprice, line 117: We should be more specific here. Something to the effect of "automatically open browser to the specified `host:port` when starting the dev server..."

It's the `host:port` piece that's helpful to clarify. Make sense?

docs/appConfiguration.md
```diff
@@ -1,47 +1,143 @@
...
+    path: './api',
+    target: TargetEnum.NODE,
+  },
+  browser: {
+    open: true,
+  },
+}
+```
+
+The options and their structure are based on Redwood's notion of sides and targets. Right now, Redwood has two fixed sides, api and web, that target NodeJS Lambdas and Browsers respectively. In the future, we'll add support for more sides and targets, like Electron and React Native (you can already see them listed as enums in [TargetEnum](https://github.com/redwoodjs/redwood/blob/d51ade08118c17459cebcdb496197ea52485364a/packages/internal/src/config.ts#L11-L12)), and as we do, you'll see them reflected in `redwood.toml`. But right now, you'll most likely never touch options like `target`.
```
thedavidprice, line 42: **Editorial style question/decision re: "Redwood S(s)ides"**
When used in the context of specifying the Redwood concept, I've been capitalizing "Side" in an attempt to clarify this is something important/specific Redwood, e.g. it's a proper noun in the Redwoodverse™.

@peterp @jtoar what are your thoughts about deciding one way or the other? I'll defer to you both. I just think we need to choose one way to handle consistently.

Also, don't let this discussion hold up merging this doc. If it needs more time, please merge this and then handle in separate issue. Thanks!

thedavidprice:
@jtoar This looks great to me and you have my 👍 to merge as is at your convenience. (I'm fine with the web table and did view it from small to x-large size.)

I left two comments you may choose to address now or after merging.

Thanks again!!

docs/appConfiguration.md
jtoar: Totally. Also saw the discussion you’re referring to. I’ll reference it when making this change.

docs/appConfiguration.md
jtoar: I’ll have to think about this one a little more/try it out. I was just reading the webpack docs, which has so many specific terms, and I can see how something like this would’ve been helpful. 

But this makes me think that (regardless of our decision here) we should add a Glossary.

jtoar:
@thedavidprice Thanks for the comments. In the spirit of more modular PRs, I’ll merge and address separately.

docs/appConfiguration.md
jtoar: @thedavidprice I'm warming up to the idea of capitalizing. Working on the testing doc now and capitalizing "Services" seems kind of natural.

jtoar: We actually already do it on the home page:

rw-capital

PR #175, opened 2020-06-07
Disable turbolinks for anchor tags in the "On this page" nav

### LINKED ISSUE #169: Links in the "On this page" nav are (kind of) broken
The links in the "On this page" nav seem kind of broken. Easiest to see with the [CLI Commands](https://redwoodjs.com/docs/cli-commands) page. The behavior I'm seeing is that when the url is 

```
https://redwoodjs.com/docs/cli-commands
```

You can use any link in the on "On this page" nav once, but then all other links take you back to the top/just don't work.

Can't point to exactly how yet but things seem to have broken with [dea43619d7d6a5b7afc415626bc381073b6a5f4f](https://github.com/redwoodjs/redwoodjs.com/commit/dea43619d7d6a5b7afc415626bc381073b6a5f4f); checking-out the one before it gives the expected behavior. I'll try to diagnose.

jtoar:
This PR closes #169 by disabling turbolinks for anchor tags in the "On this page" nav.

The nav_controller was disconnecting on every time we clicked a link in "On this page" nav, causing the custom navigate event to be dispatched and the window to scroll the top.

@cannikin I know next to nothing about turbolinks so let me know if we shouldn't be doing this!

### DIFF
--- a/lib/templates/page.html.template
+++ b/lib/templates/page.html.template
@@ -10,7 +10,7 @@
         <ul class="mt-4 pointer-events-auto">
           <% links.forEach(link => { %>
           <li class="my-1 py-1" style="margin-left: ${link.ml}rem;">
-            <a href="${link.href}">${link.title}</a>
+            <a href="${link.href}" data-turbolinks="false">${link.title}</a>
           </li>
           <% }) %>
         </ul>


cannikin:
Thanks! I've been using Turbolinks for years and never noticed this behavior...apparently it's a real thing! https://bloggie.io/@kinopyo/prevent-turbolinks-from-reloading-the-page-when-clicking-anchor-links

PR #176, opened 2020-06-07
Add information about loaders and sass.

peterp:
Deploy when `0.9.0` is shipped.


jtoar:
@peterp Thanks for this, the previous wording made it sound like either option was overriding the whole base config.

thedavidprice:
@jtoar This PR just reminds me we need to figure out general documentation for CSS. I think that's on one of your amazing lists somewhere, but the information linking Webpack and SASS/SCSS should be included when/if we get there.

PR #194, opened 2020-06-14
_nav.html: Don't include /index.html in href

### LINKED ISSUE #188: "Open link in new tab" gives a 404
Trying to open any of the top-level nav items in the left navigation bar on redwoodjs.com in a new tab or window gives you a 404 page.

image

image

Tobbe:
Fixes #188


Tobbe:
Instead of adding several more redirect rules, I just removed the "index.html" part of the links. So now it will reuse the redirects we already have defined

cannikin:
> Instead of adding several more redirect rules, I just removed the "index.html" part of the links. So now it will reuse the redirects we already have defined

That's a much better plan than mine, I was thinking we had to add a bunch of annoying redirects!

PR #195, opened 2020-06-14
Update version strings

Tobbe:
It's less confusing for new contributors, and less clutter for everyone in their `git status`, `git diff`, etc output if the version strings are kept up to date in the committed code.


thedavidprice:
@cannikin is the `"version": "v0.10.0"` even being used now per page? I didn't see it in the header meta... but what do I know ¯\_(ツ)_/¯

cannikin:
The build script swaps out this number on the fly each time, I just haven't been checking the change into the repo each time.

cannikin:
Thanks @Tobbe !

PR #196, opened 2020-06-14
package.json: Explicity use yarn to make this work on Windows

Tobbe:
Environment variables, like `$npm_execpath` are referenced using a different syntax on Windows. Windows uses `%npm_execpath%`. 

Even if we got the value of that env var, it's just a path to a .js file. And those aren't executable on Windows, so we would have to prefix it with `node `. 

And, if we did that, it would still only work with yarn, because `$npm_execpath watch` would have to be `$npm_execpath run watch` to work with npm

Explicitly just using `yarn` instead of that env var is easier to understand, and works on both Windows and Linux/Mac

(If we really want to support both npm and yarn I'd be happy to update this PR to use `cross-env` instead, and exploit the fact that yarn also allows the usage of `run`, which for npm is mandatory)


thedavidprice:
Ha! I can definitely +1 @Tobbe's reasoning here with npm + env vars + cross-platform support. I just didn't know if you had a specific need for npm. If not, then definitely do the merge thingy.

😉

PR #202, opened 2020-06-17
Add Cells doc

jtoar:
> Link to cells doc: https://deploy-preview-202--redwoodjs.netlify.app/docs/cells
> Original issue: https://github.com/redwoodjs/redwood/issues/585.

This PR adds a doc on cells to redwoodjs.com. Still very much a WIP; a lot left to do here. But a question this could've answered came up recently in Discord so felt like I really should get going.

### DIFF
--- a/docs/cells.md
+++ b/docs/cells.md
@@ -245,7 +245,7 @@ client.query({
 ### Can I Perform a Mutation in a Cell?
 
 Absolutely. We do so in our [example todo app](https://github.com/redwoodjs/example-todo/blob/f29069c9dc89fa3734c6f99816442e14dc73dbf7/web/src/components/TodoListCell/TodoListCell.js#L26-L44).
-We also don't think it's an anti-pattern to do so. Far from it--your cells might end up containing a lot of logic and really serve as the hub of your app in many ways.
+We also don't think it's an anti-pattern to do so. Far from it—your cells might end up containing a lot of logic and really serve as the hub of your app in many ways.
 
 It's also important to remember that, besides exporting certain things with certain names, there aren't many rules around Cells&mdash;everything you can do in a regular component still goes.
 
--- a/docs/cells.md
+++ b/docs/cells.md
@@ -1,0 +1,368 @@
+# Cells
+
+Cells are a declarative approach to data fetching and one of Redwood's signature modes of abstraction. In a way, Cells create space: by providing conventions around data fetching, Redwood can get in between the request and the response and perform optimizations, all without you ever having to change your code.
+
+While it might seem like there must be lot of magic involved, a Cell is actually just a [higher-order component](https://reactjs.org/docs/higher-order-components.html) that executes a GraphQL query and manages its lifecycle.
+All the logic's actually in just one file: [withCell.js](https://github.com/redwoodjs/redwood/blob/master/packages/web/src/graphql/withCell.js).
+The idea is that, by exporting named constants that match the parameters of `withCell`, Redwood can assemble this higher-order component out of these constants at build-time using a babel plugin!
+
+All of this without writing a line of imperative code. Just say what is supposed to happen when, and Redwood will take care of the rest.
+
+## Generating a Cell
+
+You can generate a Cell with:
+
+```terminal
+yarn rw g cell <name>
+```
+
+This creates the directory named `<name>Cell` in `web/src/components` with four files:
+
+```terminal
+~/redwood-app$ yarn rw g cell user
+yarn run v1.22.4
+$ /redwood-app/node_modules/.bin/rw g cell user
+  ✔ Generating cell files...
+    ✔ Writing `./web/src/components/UserCell/UserCell.mock.js`...
+    ✔ Writing `./web/src/components/UserCell/UserCell.stories.js`...
+    ✔ Writing `./web/src/components/UserCell/UserCell.test.js`...
+    ✔ Writing `./web/src/components/UserCell/UserCell.js`...
+Done in 1.07s.
+```
+
+We'll go over each of these files in detail. But know that the file appended with just `.js` (in the example above, `UserCell.js`) contains all your Cell's logic.
+
+Off the bat, this file exports five constants: `QUERY`, `Loading` , `Empty` , `Failure`  and `Success`. The root query in `QUERY` is the same as `<name>` so that, if you're generating a cell based on a model in your `schema.prisma`, you can get something out of the database right away. But there's a good chance you won't generate your Cell this way, so if you need to, make sure to change the root query. See the [Cells](https://deploy-preview-202--redwoodjs.netlify.app/tutorial/cells#our-first-cell) section of the Tutorial for a great example of this.
+
+## Usage
+
+With Cells, you have a total of seven exports to work with:
+
+| Name          | Type              | Description                                                  |
+| :------------ | :---------------- | :----------------------------------------------------------- |
+| `QUERY`       | `string|function` | The query to execute                                         |
+| `beforeQuery` | `function`        | Lifecycle hook; prepares variables and options for the query |
+| `afterQuery`  | `function`        | Lifecycle hook; sanitizes data returned from the query       |
+| `Loading`     | `component`       | If the request is in flight, render this component           |
+| `Empty`       | `component`       | If there's no data (`null` or `[]`), render this component   |
+| `Failure`     | `component`       | If something went wrong, render this component               |
+| `Success`     | `component`       | If the data has loaded, render this component                |
+
+Only `QUERY` and `Success` are required. If you don't export `Empty`, empty results are sent to `Success`, and if you don't export `Failure`, error is output to the console.
+
+`Loading`, `Empty`, `Failure`, and `Success` all have access to the same set of props, with `Failure` and `Success` getting exclusive access to `error` and `data` respectively. So, in addition to displaying the right component, a Cell funnels the right props to the right component.
+
+This prop set is composed of 1) what's returned from Apollo Client's `Query` component, which is quite a few things&mdash;see their [API reference](https://www.apollographql.com/docs/react/api/react-components/#render-prop-function) for the full list (note that, as we just mentioned, `error` and `data` are only available to `Failure` and `Success` respectively. And Cells use `loading` to decide when to show `Loading`, so you don't get that one either)&mdash;and 2) props passed down from the parent component in good ol' React fashion.
+
+### QUERY
+
+`QUERY` can be a string or a function. Note that it's totally more than ok to have more than one root query. Here's an example of that:
+
+```javascript{7-10}
+export const QUERY = gql`{
+  query {
+    posts {
+      id
+      title
+    }
+    authors {
+      id
+      name
+    }
+  }
+}
+```
+
+So in this case, both `posts` and `authors` would be available to `Success`:
+
+```js
+export const Success = ({ posts, authors }) => {
+  // render logic with posts and authors
+}
+```
+
+If `QUERY` is a function, it has to return a valid GraphQL syntax tree.
+Use a function if your queries need to be more dynamic:
+
+<!-- Source: https://community.redwoodjs.com/t/custom-github-jwt-auth-with-redwood-auth-advice-needed/610 -->
+But what about variables? Well, Cells are setup to use any props they receive from their parent as variables (things are setup this way in `beforeQuery`). For example, here `BlogPostCell` takes a prop, `numberToShow`, so `numberToShow` is just available to your `QUERY`:
+
+```javascript{7}
+import BlogPostsCell from 'src/components/BlogPostsCell'
+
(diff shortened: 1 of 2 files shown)


docs/cells.md
```diff
@@ -0,0 +1,152 @@
+# Cells
+
+<!-- JSparty 38:00 -->
+
+Cells are a declarative approach to data fetching and one of Redwood's signature modes of abstraction.
+
+In a way, cells create space: Redwood can get in between the request and the response and perform optimizations, all without you ever having to change your code.
+
+There must be a lot of magic involved right? Well, maybe a little less than you'd expect: a cell's actually just a [higher-order component](https://reactjs.org/docs/higher-order-components.html) that executes a GraphQL query (here, the aptly-named `QUERY`) and manages its lifecycle.
```
Tobbe, line 9:
```suggestion
There must be a lot of magic involved right? Well, maybe a little less than you'd expect: a cell is actually just a [higher-order component](https://reactjs.org/docs/higher-order-components.html) that executes a GraphQL query (here, the aptly-named `QUERY`) and manages its lifecycle.
```

We've talked about this before I think, but I'll bring it up again. I know you like the contraction because it makes it less formal, but to me this looks too much like the possessive form, and it throws me off when reading.

Just putting my 2c in there, keep it as it is if you want :)

docs/cells.md
```diff
@@ -0,0 +1,152 @@
...
+| :------------ | :---------------- | :--------------------------------------------- |
+| `QUERY`       | `string|function` | The GraphQL syntax tree to execute             |
+| `beforeQuery` | `function`        | Prepares variables and options for the query   |
+| `afterQuery`  | `function`        | Sanitizes data returned from the QUERY         |
+| `Loading`     | `component`       | If loading, render this component              |
+| `Empty`       | `component`       | If there's no data, render this component      |
+| `Failure`     | `component`       | If something went wrong, render this component |
+| `Success`     | `component`       | If the data has loaded, render this component  |
+
+Only `QUERY` and `Success` are rquired. If you don't export `Empty`, empty results will be sent to `Success`, and if you don't export `Failure`, error output will be sent to the console.
```
Tobbe, line 78:
```suggestion
Only `QUERY` and `Success` are required. If you don't export `Empty`, empty results will be sent to `Success`, and if you don't export `Failure`, error output will be sent to the console.
```

docs/cells.md
```diff
@@ -0,0 +1,152 @@
...
+
+```javascript
+export const beforeQuery = (props) => {
+  return { variables: props, fetchPolicy: 'network-only' }
+}
+```
+
+See [Customizing the Fetch Policy](https://www.apollographql.com/docs/tutorial/queries/#customizing-the-fetch-policy).
+
+`QUERY` can be a string or a function (if you need to write dynamics queries).
```
Tobbe, line 91:
```suggestion
`QUERY` can be a string or a function (if you need to write dynamic queries).
```

docs/cells.md
```diff
@@ -0,0 +1,152 @@
...
+      }
+    }
+    ${EntityList}
+  `
+}
+```
+
+### When should I use a cell?
+
+A good rule of thumb for when to use cells is if your component needs some data from the database or other service that may be delayed in responding. Let Redwood worry about juggling what is displayed when and you can focus on the happy path of the final, rendered component populated with data.
```
Tobbe, line 111:
```suggestion
A good rule of thumb for when to use cells is if your component needs some data from the database or other service that may be delayed in responding. Let Redwood worry about juggling what is displayed when and you can focus on the happy path of the final, rendered, component populated with data.
```

docs/cells.md
```diff
@@ -0,0 +1,152 @@
...
+
+<!-- Source: https://github.com/redwoodjs/redwood/pull/413 -->
+For one-off queries, you can always use `useApolloClient`, which returns the client. For all the functions on th client:
```
Tobbe, line 114:
```suggestion
For one-off queries, you can always use `useApolloClient`, which returns the client. For all the functions on the client, see:
```

docs/cells.md
```diff
@@ -0,0 +1,152 @@
...
+| `afterQuery`  | `function`        | Sanitizes data returned from the QUERY         |
+| `Loading`     | `component`       | If loading, render this component              |
+| `Empty`       | `component`       | If there's no data, render this component      |
+| `Failure`     | `component`       | If something went wrong, render this component |
+| `Success`     | `component`       | If the data has loaded, render this component  |
+
+Only `QUERY` and `Success` are rquired. If you don't export `Empty`, empty results will be sent to `Success`, and if you don't export `Failure`, error output will be sent to the console.
+
+<!-- Source: https://github.com/redwoodjs/redwood/issues/717 -->
+beforeQuery has a default `fetchPolicy` of `'cache-and-network'`. But you can change this by exporintg your own:
```
Tobbe, line 81:
```suggestion
beforeQuery has a default `fetchPolicy` of `'cache-and-network'`. But you can change this by exporting your own:
```

docs/cells.md
```diff
@@ -0,0 +1,152 @@
...
+All the logic's actually just in one file: [withCell.js](https://github.com/redwoodjs/redwood/blob/master/packages/web/src/graphql/withCell.js).
+The idea is: by exporting named constants that match the parameters of `withCell`, Redwood can assemble this higher-order component out of these constants at build-time using a babel-plugin!
+
+When React renders this higher-order component, Redwood performs the `QUERY` and displays the `Loading` component till the query returns. Once it does, it'll display one of three components: `Failure` if there was an error; `Empty` if the data return is null or an empty array; and `Success` otherwise. -->
+
+## How does Redwood know a cell is a cell?
+
+<!-- Source: https://github.com/redwoodjs/redwood/pull/597 -->
+<!-- Source: https://github.com/redwoodjs/redwood/pull/554 -->
+You just have to add Cell to the end right? Well, while that was actually the case, things are a little more sophisiticated now: [todo].
```
Tobbe, line 144:
```suggestion
You just have to add Cell to the end right? Well, while that was actually the case, things are a little more sophisticated now: [todo].
```

docs/cells.md
```diff
@@ -0,0 +1,152 @@
...
+```javascript
+export const beforeQuery = (props) => {
+  return { variables: props, fetchPolicy: 'network-only' }
+}
+```
+
+See [Customizing the Fetch Policy](https://www.apollographql.com/docs/tutorial/queries/#customizing-the-fetch-policy).
+
+`QUERY` can be a string or a function (if you need to write dynamics queries).
+If it's a function, it has to return a valid GraphQL syntax tree:
```
Tobbe, line 92: Would it be possible to add something about what parameters would be passed to this function? I see `model` and `singular` in the example below, but I don't know if those are the only things that are passed, or if there are more. And I'm also missing a description of what they are.

docs/cells.md
```diff
@@ -0,0 +1,152 @@
...
+There's a total of seven exports to work with:
+
+| Name          | Type              | Description                                    |
+| :------------ | :---------------- | :--------------------------------------------- |
+| `QUERY`       | `string|function` | The GraphQL syntax tree to execute             |
+| `beforeQuery` | `function`        | Prepares variables and options for the query   |
+| `afterQuery`  | `function`        | Sanitizes data returned from the QUERY         |
+| `Loading`     | `component`       | If loading, render this component              |
+| `Empty`       | `component`       | If there's no data, render this component      |
+| `Failure`     | `component`       | If something went wrong, render this component |
```
Tobbe, line 75: It would be awesome if you could (somewhere) explain what props these components will receive. I just had a use case where I did this:
```
export const Failure = ({ error, variables }) => {
```
I did that to be able to show a specific error message for a specific variable value

docs/cells.md
jtoar: Totally, trying to lose the habit with non-pronouns!

docs/cells.md
jtoar: Absolutely, was using the example more as a placeholder. I pulled it from the original PR.

docs/cells.md
jtoar: Definitely, this is a must.

docs/cells.md
jtoar: Thinking I should just rework some of this sentence altogether.

thedavidprice:
**re: adding an example of using a Function for a Query from #505**

@eurobob could you take a look at [this section](https://deploy-preview-202--redwoodjs.netlify.app/docs/cells#query) of the new Cell doc in progress -- possible you could help complete the example and explanation for using a function as a Query? Thanks in advance!

jtoar:
Note to self: add a section about `useQuery` being available.

dthyresson:
I added an issue https://github.com/redwoodjs/redwoodjs.com/issues/307 as well to cover some potential additional topics.

docs/cells.md
```diff
@@ -0,0 +1,368 @@
+# Cells
+
+Cells are a declarative approach to data fetching and one of Redwood's signature modes of abstraction. In a way, Cells create space: by providing conventions around data fetching, Redwood can get in between the request and the response and perform optimizations, all without you ever having to change your code.
+
+While it might seem like there must be lot of magic involved, a Cell is actually just a [higher-order component](https://reactjs.org/docs/higher-order-components.html) that executes a GraphQL query and manages its lifecycle.
+All the logic's actually in just one file: [withCell.js](https://github.com/redwoodjs/redwood/blob/master/packages/web/src/graphql/withCell.js).
```
thedavidprice, line 6:
```suggestion
All the logic's actually in just one file: [withCell.js](https://github.com/redwoodjs/redwood/blob/main/packages/web/src/graphql/withCell.js).
```

docs/cells.md
```diff
@@ -0,0 +1,368 @@
...
+    ✔ Writing `./web/src/components/UserCell/UserCell.mock.js`...
+    ✔ Writing `./web/src/components/UserCell/UserCell.stories.js`...
+    ✔ Writing `./web/src/components/UserCell/UserCell.test.js`...
+    ✔ Writing `./web/src/components/UserCell/UserCell.js`...
+Done in 1.07s.
+```
+
+We'll go over each of these files in detail. But know that the file appended with just `.js` (in the example above, `UserCell.js`) contains all your Cell's logic.
+
+Off the bat, this file exports five constants: `QUERY`, `Loading` , `Empty` , `Failure`  and `Success`. The root query in `QUERY` is the same as `<name>` so that, if you're generating a cell based on a model in your `schema.prisma`, you can get something out of the database right away. But there's a good chance you won't generate your Cell this way, so if you need to, make sure to change the root query. See the [Cells](https://deploy-preview-202--redwoodjs.netlify.app/tutorial/cells#our-first-cell) section of the Tutorial for a great example of this.
```
thedavidprice, line 35:
```suggestion
Off the bat, this file exports five constants: `QUERY`, `Loading` , `Empty` , `Failure`  and `Success`. The root query in `QUERY` is the same as `<name>` so that, if you're generating a cell based on a model in your `schema.prisma`, you can get something out of the database right away. But there's a good chance you won't generate your Cell this way, so if you need to, make sure to change the root query. See the [Cells](https://redwoodjs.com/tutorial/cells#our-first-cell) section of the Tutorial for a great example of this.
```

docs/cells.md
```diff
@@ -0,0 +1,368 @@
...
+  query: gql`
+    ...
+  `
+})
+```
+
+### Can I Perform a Mutation in a Cell?
+
+Absolutely. We do so in our [example todo app](https://github.com/redwoodjs/example-todo/blob/f29069c9dc89fa3734c6f99816442e14dc73dbf7/web/src/components/TodoListCell/TodoListCell.js#L26-L44).
+We also don't think it's an anti-pattern to do so. Far from it--your cells might end up containing a lot of logic and really serve as the hub of your app in many ways.
```
thedavidprice, line 248:
```suggestion
We also don't think it's an anti-pattern to do so. Far from it—your cells might end up containing a lot of logic and really serve as the hub of your app in many ways.
```

docs/cells.md
```diff
@@ -0,0 +1,368 @@
...
+
+It's also important to remember that, besides exporting certain things with certain names, there aren't many rules around Cells&mdash;everything you can do in a regular component still goes.
+
+## How Does Redwood Know a Cell is a Cell?
+
+You just have to end a filename in "Cell" right? Well, while that's basically correct, there is one other thing you should know.
+
+Redwood looks for all files ending in "Cell" (so if you want your component to be a Cell, its filename does have to end in "Cell"), but if the file 1) doesn't export a const named `QUERY` and 2) has a default export, then it'll be skipped.
+
+When would you want do this? If you just want a file to end in "Cell" for some reason. Otherwise, don't worry about it!
```
thedavidprice, line 258:
```suggestion
When would you want to do this? If you just want a file to end in "Cell" for some reason. Otherwise, don't worry about it!
```

thedavidprice:
This is just amazing @jtoar Thank you 🚀

PR #206, opened 2020-06-19
Update `db save` tutorial notes

Tobbe:
As of this https://github.com/redwoodjs/redwood/pull/708 the Windows warning in the tutorial is not needed anymore


cannikin:
Good catch, thanks!

PR #208, opened 2020-06-19
Tutorial: Windows note about rw g page home /

Tobbe:
I was bit by this today again when I did a quick run-through of the tutorial. So now I added a warning ⚠
https://github.com/redwoodjs/redwood/issues/574

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -170,7 +170,11 @@ Let's give our users something to look at besides the Redwood welcome page. We'l
 
     yarn redwood generate page home /
 
-This does three things:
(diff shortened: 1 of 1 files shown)


Tobbe:
I had a longer text first, but decided to keep it shorter. The longer text (obviously) had a bit more details/info. Let me know if you want me to use that one instead, or something in-between

> ⚠️ **Windows Users:** there is a known Git for Windows issue causing the `yarn redwood generate page home /` command to generate the wrong path for the route in `web/src/Routes.js`. Thankfully there's an easy workaround. Just run the command with double slashes; `yarn redwood generate page home //`.
>
> If you had already run the command before seeing this warning, you can manually go in and edit the `web/src/Routes.js` file. The `path` should be `path="/"`.
>
> You can check the current status of this issue [here](https://github.com/redwoodjs/redwood/issues/574). Apologies for the inconvenience!

PR #210, opened 2020-06-21
Fix a typo in postgres word

amrrkf:
Fix a typo in a `postgres` word. I found the typo in the guide for local development using Postgresql.


cannikin:
Good catch, thanks!

PR #217, opened 2020-06-28
Spike Glossary

jtoar:
> [Link to glossary](https://deploy-preview-217--redwoodjs.netlify.app/docs/glossary)

In #173, @thedavidprice brought up the idea of clarifying proper nouns in the Redwoodverse™ by capitalizing them. Whether or not we go that route, to the same end, we should consider adding a glossary.

Webpack is a great example. They have so many specific terms. A webpack module is not a JavaScript module (and it helps to know that sooner than later). While I think we do a great job of explaining Redwood terms in the tutorial, a glossary is a quick, convenient way to look up/clarify terms users already kind of know but want to be sure.

This is very much a WIP, and the terms listed right now are the ones that are obvious, with their definitions being more of TODO-placeholders than final cuts. Any suggestions/definitions would be greatly appreciated! I also think we should define some Jamstack specific terms as there's a lot of overlap.

### DIFF
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -24,13 +24,13 @@ TODO
 
 ## Services
 
-Redwood's opinionated way of organizing business logic code 
+Redwood's opinionated way of organizing your business logic for maintainability.
 
-## side
+## Side
 
 TODO
 
-## target
+## Target
 
 TODO
 
--- a/docs/glossary.md
+++ b/docs/glossary.md
@@ -4,7 +4,7 @@ Redwood and Jamstack terms defined. Apart from being a great reference, glossari
 
 ## Cell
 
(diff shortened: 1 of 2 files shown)


docs/glossary.md
```diff
@@ -0,0 +1,40 @@
+# Glossary
+
+Redwood and Jamstack terms defined. Apart from being a great reference, glossaries are a great way to test your knowledge. Do you know each one of these terms?
+
+## Cell
+
+Redwood's simpler and more decalarative approach to data fetching.
```
yg, line 7: Typo

```suggestion
Redwood's simpler and more declarative approach to data fetching.
```

jtoar:
Not convinced we need this yet; closing for now.

PR #218, opened 2020-06-28
Fix typo in app configuration, update version numbers

dfundingsland:
This PR fixes a typo in the api configuration table where the path to the api side is given as `'./web'` as well as some outdated version references.


jtoar:
@dfundingsland Thanks so much for this!

PR #230, opened 2020-07-05
Reorganize Videos section + proposal

jtoar:
> [Link to new deploy](https://deploy-preview-230--redwoodjs.netlify.app/). Click **Videos** in the nav. It should toggle now. Then try navigating to **Videos** > **Tutorial**.

This is a housekeeping PR that reorganizes the **Videos** section a bit, making adding videos more like adding docs/cookbook recipes&mdash;just put a file in a directory.

But we should do more. [Via the forum](https://community.redwoodjs.com/t/basic-question-about-relations/770/5):

> I know it is easier said than done but some video tutorials and cookbooks focused on services would be super useful in terms of showcasing their true potential and to illustrate to a casual observer how they can truly make your Redwood app full stack. **Undoubtedly, in my opinion, sites like Laracasts have really helped grow the Laravel community**. I know it’s early days, but something akin to that could be amazing for the adoption curve of Redwood.

The part here that I really agree with is the adoption curve. Videos === adoption? There's some meaningful correlation there. And the best part is, we already have amazing videos! In both mediums, the Tutorial is masterful.

But [right now](https://redwoodjs.com/videos), the videos section is a long-form scroll with no nav. This PR divides it up into sections and restores the nav, just to get us started.

I don't think we should go as far as [Laracasts](https://laracasts.com/), where you login to your spaceship dashboard, etc. But the least we could do is design it more like [Livewire](https://laravel-livewire.com/screencasts/installation).


cannikin:
Nice! Only weird thing I saw was on the router page:

image

jtoar:
@cannikin Was waiting for that one to be uploaded to youtube so we could just link there&mdash;then the ## section would make more sense, but I'll remove it for now.

jtoar:
@cannikin Woops didn't realize what was actually going on there. Thought the section title was being repeated by the tweet embed, but my eyes just glazed over the html.

PR #231, opened 2020-07-05
Edit Tutorial sections Prereqs and Wrapping Up

jtoar:
This PR applies 2/3 of the edits Rob mentioned in the forum topic [Grokking Redwood](https://community.redwoodjs.com/t/groking-redwood/772/2) (reprinting Rob's comments for context):

- [x] Add Prisma to the list in Prerequisites. I also added Apollo. 

> Yeah I mean we specifically state at the start of the tutorial that you should be familiar with React and GraphQL (maybe we could add Prisma to that list?) ...

- [x] Add another paragraph to the Wrapping Up section of the Tutorial for users ready to take the next step.

> We could also link to specific pages in the Wrapping Up section of the tutorial for users who are ready to take the next step and include links to Apollo and Prisma docs, and link to specific docs depending on your familiarity with various concepts (never used GraphQL before, just need an overview of writing mutations, etc.)

- [ ] Clarify/expand on Prisma's place in the data flow

> We (Tom and I) felt that the [Side Quest: How Redwood Works with Data](https://redwoodjs.com/tutorial/side-quest-how-redwood-works-with-data) section was a good spot for an interlude and is hopefully just enough theory to help those that are curious about where this data is coming from and where it’s going. Although looking through it now I see it only mentions Prisma once, in the little sidebar about the db variable. I’m all for clarifying that data flow in this section.


thedavidprice:
I could be fine with adding these as-is, but what about being a bit more informative, e.g.:
- [React](https://reactjs.org/) for building the UI 
- [GraphQL](https://graphql.org/) API query language using the [Apollo](https://www.apollographql.com/docs/tutorial/introduction/) implementation
- [Prisma](https://www.prisma.io/) for database access (query builder, connection, and migrations)
- [The Jamstack](https://jamstack.org/) methodology and infrastructure

**re: Prisma**
These are the two Prisma docs I think it would be most helpful to refer:
- Schema: https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-schema
- CRUD: https://www.prisma.io/docs/reference/tools-and-interfaces/prisma-client/crud

cannikin:
Not sure if this PR still even applies?? But either way, we've moved all of the tutorials over to the https://learn.redwoodjs.com site! We'll need to make any future changes to the tutorial(s) over there instead: https://github.com/redwoodjs/learn.redwoodjs.com

Issue #181, opened 2020-06-08
Add environment variable to disable search indexing in development


cannikin:
When people are building the site locally they need to comment out some code in order to skip running the search index (which will fail without an Algolia api key). Let's just set an ENV var that says whether or not to worry about populating the search during build time, set to `false` by default—usually you don't care about the search index during development.

We will be sure to set the flag ON for Netlify deployments.

jtoar:
I think #324 addressed this; let me know if not, but I've been able to develop locally without worrying about search since.

PR #232, opened 2020-07-06
Add Prisma primer

jtoar:
WIP Prisma primer in response to the [Grokking Redwood](https://community.redwoodjs.com/t/groking-redwood/772) topic on the forum.

My working theory of difficulty, as someone who didn't really know TypeScript or Prisma (but now kind of does, or at least how they're related): we have to point users to `node_modules/.prisma/client`&mdash;the directory where Prisma Client lives, because the `index.d.ts` file there basically is your docs for your Prisma Client.

That or my intellisense was just never working.

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -110,6 +110,10 @@ const SECTIONS = [
       },
       {
         pageBreakAtHeadingDepth: [1],
+        url: './docs/prisma.md',
+      },
+      {
+        pageBreakAtHeadingDepth: [1],
         url: 'redwoodjs/redwood/packages/router/README.md',
       },
       {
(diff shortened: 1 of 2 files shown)


lib/build.js
```diff
@@ -10,7 +10,7 @@ const octokit = Octokit({
 })
 
 const { create: createDocs, buildNav } = require('./docutron.js')
-const { publish: publishSearch, getObjectIDs } = require('./search.js')
+// const { publish: publishSearch, getObjectIDs } = require('./search.js')
```
cannikin, line 13: Careful, the commented out search indexing is actually committed!

jtoar: Thanks! 🤦‍♂️ last time I’ll make this mistake

cannikin: No problem! I've got an issue #181 to just add an ENV var to prevent having to do this going forward!

Issue #183, opened 2020-06-09
Add missing <SelectField> to Form Doc


thedavidprice:
See `<SelectField>` here in `form.sj`:
https://github.com/redwoodjs/redwood/blob/d51ade08118c17459cebcdb496197ea52485364a/packages/web/src/form/form.js#L210

We should also audit to see if any other components/elements are missing from doc.

cc @jtoar

jtoar:
Closed by #235; @tctrautman also checked to see if any others were missing and there weren't.

PR #235, opened 2020-07-07
Add <SelectField /> to docs

tctrautman:
This is a tiny PR to add `<SelectField>` to the list of Form components, fixing https://github.com/redwoodjs/redwoodjs.com/issues/183.  And, per the request in the issue, I double checked to make sure there aren't other Form components missing in the docs and didn't see anything else missing.

Also it's worth noting that, upon trying to install and build the site locally, I was able to view the homepage just fine, but shown a "page not found" error when I tried to navigate to the `/docs/form` page.  I assume this has more to do with my dev environment (running node 10.15.3 by default) than it does with the project or my small markdown edit.

### DIFF
--- a/docs/form.md
+++ b/docs/form.md
@@ -9,6 +9,7 @@ Redwood currently provides the following form components:
 * `<Form>` surrounds all form elements and provides contexts for errors and form submission
 * `<FormError>` displays an error message, typically at the top of your form, containing error messages from the server
 * `<Label>` is used in place of the HTML `<label>` tag and can respond to errors with different styling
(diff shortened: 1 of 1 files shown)


cannikin:
> I was able to view the homepage just fine, but shown a "page not found" error when I tried to navigate to the `/docs/form` page

Yeah the dev server doesn't know how to follow redirects defined in `netlify.toml`. You could try `netlify dev` (need to install `netlify-cli` globally) but I think there may have been issues with that as well...

jtoar:
@tctrautman Thanks for adding this and for checking to see if any others were missing, that's really going the extra mile!

PR #238, opened 2020-07-09
Change code to HTML to have coloured highlights

amorriscode:
The section of the tutorial that [deals with Routing Params](https://redwoodjs.com/tutorial/routing-params) doesn't nicely highlight one of the code sections.

image

I adjusted the `.md` file to be the same as the other examples so the nice highlighting will work! Here is a screenshot post-change:

image


jtoar:
@amorriscode Thanks! Especially for the before-after screenshots, looks much better now.

PR #244, opened 2020-07-11
Add missing route from doc example

amorriscode:
I was doing the tutorial and noticed there was a missing route during the [authentication step](https://redwoodjs.com/tutorial/authentication). It had me doing a second take so I thought I'd update it.

Feel free to close this PR if you don't think it's worth having.


jtoar:
@amorriscode Thanks! And definitely worth having, its the details that close the gap between good and great, especially when it comes to docs. This kind of consistency is super important.

PR #245, opened 2020-07-13
Add roadmap to 1.0

jtoar:
> [Link to roadmap](https://deploy-preview-245--redwoodjs.netlify.app/roadmap-to-1-0)

The PR spikes a new top-level section in the nav: "Roadmap to `1.0`". Clicking it sends you to a standalone page that has all-the-things `1.0`.

rw-roadmap

It goes without saying that I need this to be extensively cross-checked because it is, using the term as loosely as possible, a spec.

Part of the reason some of the sections are so short is I just didn't know what to say; didn't know what our plans were. So I'm hoping we can use this page as a way of getting on the same page.

Also part of the reason some of the sections are short is I intentionally kept them that way: since I set out to write an overview, I didn't want to make anything tediously long. Basically, deciding on the level of granularity was hard.

And because I wanted to provide an overview, I thought the color-coding (or at least something like it) was super necessary.

> **History:** at first I tried using Basecamp, since they've got those sweet [Hill Charts](https://basecamp.com/shapeup/3.4-chapter-12#work-is-like-a-hill)&mdash;great for overviews. But I pretty quickly just figured we'd start with this and if we want to get discussions going around specific todos, we could think about migrating there.

A few quick points:

- @cannikin Sorry for what I did to `docutron.js` I can make it less spiky
- @cannikin I pulled the icon from [zondicons](http://www.zondicons.com/) but can make one / make it look more like the others.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,0 +1,234 @@
+# Roadmap to 1.0
+
+The speed we're moving at is pretty remarkable. We want to hit `1.0` by the end of the year. And we think we can do it. But there are a lot of moving parts, each at a different level of maturity. 
+
+At this stage of development, when it's so important to keep the finish line in mind, having a high-level overview is invaluable. Hence, this roadmap, and these color-coded labels:
+
+- <span id="status-0" class="font-mono">Didn't start</span>
+- <span id="status-1" class="font-mono">Figuring it out</span>
+- <span id="status-2" class="font-mono">There's a plan</span>
+- <span id="status-3" class="font-mono">Making it happen</span>
+- <span id="status-4" class="font-mono">Cleaning up</span>
+
+We're taking inspiration from Basecamp's [Shape Up](https://basecamp.com/shapeup/3.4-chapter-12#work-is-like-a-hill). You can think of each one of these color-coded labels as a point on the hill chart.
+
+> If you know Shape Up then you'll know that our scopes are probably too big. They are, but we want to orient people who aren't knee-deep in this, and questions like "how's TypeScript going?" are usually the kinds they have. So even though TypeScript is a big scope that comprises a lot of smaller, more actionable scopes, we want to give you an idea of where we're at with it.
+> 
+> This means that even if we say TypeScript is <span id="status-3" class="font-mono">Making it happen</span>, there are some parts of it that might be <span id="status-0" class="font-mono">Didn't start</span>.
+
+If you want an even higher-level overview, scroll the aside on the right, keeping the color key in mind.
+
+This document is alive. We'll update it as often as we can, and when it's appropriate to do so. And as always, feel free to open a PR!
+
+## What goes in 1.0?
+
+Not everything will be in `1.0`. Not everything can go in `1.0`. Even some really great ideas won't be&mdash;even things that are core to the Redwood dream, like more sides and targets (because those will be in [`2.0`](#20)!). 
+
+We need to be very careful about our priorities, both because we want `1.0` to be something special and because if we don't be careful, we'll never get there. The hardest thing in open source is saying no. But as we get closer to `1.0`, it's increasingly what we'll have to do.
+
+Just to let you know that we haven't forgotten about some topics and have them in our long-term plans, there's a section at the bottom: [`2.0`](#20). You're more than welcome to work on one of these.
+
+But for `1.0`, these are all the must-haves:
+
+## Accessibility
+
+<span id="status-1" class="font-mono">Figuring it out</span>
+
+Accessibility isn't something we're going to compromise on. 
+It has to be first class. 
+It's in the work at [#540](https://github.com/redwoodjs/redwood/issues/540).
+Accessibility mainly plays through the Router.
+As background reading, Gatsby has some [great blog posts](https://www.gatsbyjs.org/blog/2020-02-10-accessible-client-side-routing-improvements/).
+
+- [Router: Focus is not reset to top of DOM on route change #540](https://github.com/redwoodjs/redwood/issues/540)
+- [[WIP] Add RouteAnnouncer for accessibility #693](https://github.com/redwoodjs/redwood/pull/693)
+
+## Auth
+
+<span id="status-2" class="font-mono">There's a plan</span>
+
+<!-- @todo RBAC; a 1.0 thing? -->
+<!-- @todo anything else -->
+
+We have big plans for Auth, and the next thing up is role-based access control (RBAC).
+RBAC is about checking to see if the current user belongs to a role that is permitted access to a Route, Page, Function, etc. It's in discussion at [#745](https://github.com/redwoodjs/redwood/issues/806).
+
+- [Simple username/password authentication, which method should I use? #745](https://github.com/redwoodjs/redwood/issues/745)
+- [Implement Role-based Authorization #806](https://github.com/redwoodjs/redwood/issues/806)
+
+## Bundle Size
+
+<span id="status-0" class="font-mono">Didn't start</span>
+
+We haven't looked. But one way you can is by building with the stats flag (`yarn rw build --stats`). This will make Redwood bundle with [Webpack Bundle Analyzer](https://github.com/webpack-contrib/webpack-bundle-analyzer).
+
+- [Possible CI solution for monitoring build size #196](https://github.com/redwoodjs/redwood/issues/196)
+- [Preact instead of React? #261](https://github.com/redwoodjs/redwood/issues/261)
+- [Use react-query for data fetching #758](https://github.com/redwoodjs/redwood/issues/758#issuecomment-650299099)
+
+## CI/CD (App-side)
+
+<span id="status-2" class="font-mono">There's a plan</span>
+
+<!-- @todo did I get that right, and what is it exactly? -->
+
+You need CI/CD nowadays if you're going to do anything real. So we'd like to provide a [GitHub Actions](https://github.com/features/actions) template you can copy.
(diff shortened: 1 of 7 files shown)


mojombo:
This is amazing! I will spend some time and make any changes/additions. Thank you so much for getting this started, I really like the structure and tags you've used to show where each component is at. 💯

cannikin:
> Sorry for what I did to docutron.js I can make it less spiky

Docutron!! That's fine with me as it...if you want to clean it up a little that's okay, but I get the gist of it!

> I used [Google Material Icons](https://material.io/resources/icons/?style=twotone) for all those, the two-tone versions. I'm there there's something very close!

jtoar:
@mojombo Anytime! Super glad you like the format. And I can totally add anything, whether it's more specific to-dos, or a tweak to the format, or just more content. I'll also be on the look out for updates.

@cannikin Sweet, thanks for that link! Yeah I'll grab something that looks way more cohesive from there.

mojombo:
The order of the status colors feels a bit odd to me. I'd put them in this order:

-_Roadmap_to_1_0___RedwoodJS_Docs

Also, the flag icon for the menu item doesn't seem to fit in with the others super well, let's see if we can track one down that lines up a little better and has a more consistent style. cc @cannikin

jtoar:
@mojombo I rearranged the order of the labels accordingly (double checking now).

Rob told me he's using the two-tone material design icons. I looked through for the ones that make the most sense. Which of these do you like more?

rw-roadmap-icons

thedavidprice:
## Existing Lists
Top of my list to spend time with this tomorrow. Couple quick things for now:
- we'll want to reconcile Roadmap with the tutorial Wrapping Up: https://redwoodjs.com/tutorial/wrapping-up
- Here's the working list I've been keeping
    - Jest Testing complete
    - Storybook integration
    - TS first-class support (framework and App)
    - Router
        - TS
        - Pre-Render
        - Accessibility
    - Deploy Targets: Google Cloud, Vercel, AWS, Azure
    - Role-based Authentication
    - Error Handling and Diagnostics
    - CI/CD hardening (framework)
    - Sides

## Possible to Group by Categories
My brain also wanted to see this with some categories versus list of things. Each category could have a general stated goal for v1 with a list of specifics. Not sure if it's possible but could be things like:
    - CI/CD & DevOps(?): Unit tests, Deployment Targets, Diagnostics and Error Handling, Logging, Package Size, Dev Server Spin Up Time, Build time
    - Architecture(?): Sides, TS/JS, Project-Structure API
    - Workflow: Storybook
    - Router
    - API
    - ...

## 2.0
I'm mixed about including this. Maybe it's less about 2.0 and more of a "here's other long-term items that belong to the full vision... e.g. serverless DB support."

> Apollo Client Alternative
I strongly recommend we _remove_ this right now. When/if we think we're serious we can start discussing publicly. But right now it might throw alarms.

jtoar:
@thedavidprice Re grouping by categories: I can add another label, or maybe make it more like the badges you see on GitHub that have the "Build | PASSING" (category | status). Also removed that bit at the end.

jtoar:
@thedavidprice Something like this: 

rw-badge

Where the category precedes the status.

thedavidprice:
@jtoar Ah, that's not what I meant to suggest but could indeed be a possibility. I was just thinking group by headers, effectively. So `<h2>CI/CD</h2> Primary v1 goal/vision for this category ... lists ...`, `<h2> Architecture </h2> Primary v1 goal/vision for this category ... lists ...`. 

But either way, before trying to implement we'd want to make sure that any kind of grouping makes sense.

Tobbe:
image

I like this order of colors better :)  But I think the previous order of status names was more logical...
So,
Didn't start - Red
Figuring it out - Yellow
There's a plan - Purple
Making it happen - Blue
Cleaning up - Green

mojombo:
@jtoar To @Tobbe's point, I intended that only the label COLORS would change to be in the order of red-yellow-purple-blue-green (not the text or order of the labels), and that change would only be a CSS change. I recommend reverting #855d54f and just changing the colors in CSS. Sorry for the confusing request!

jtoar:
@mojombo Ah got it, that makes sense. Looking at your original comment again, I see now that you specifically said colors, I glossed over that. Those changes have been made; the new color order definitely makes both sequences look like they run in parallel. Thanks!

dthyresson:
Here's my super wishlist of RedwoodJS features that I know I almost always want (and more often than not get put into the "if we have time in 1.0 when project planning). Important: RW does not necessarily have to (and probably should not) roll-own solutions, but can bring onboard those tools and wrap them, abstract out the complexity so they "just work".

## ~ 1.0 Wishlist (aka what I want in my 1.0 webapp, but usually comes in 1.2 or never)

* ✌️ Security
*  🔍  SEO
* 🗂  Prerendering
* 📝  Logging
* 🖼  Image Manipulation/Responsive Image Support
* 📈 Analytics Hooks
* 🕵️‍♀️  Search hooks
* ⚖️  Environment Parity (dev/integ/prod)
* ⚙️  3rd Party API Integration Helpers
* 🪐  The "Other Side"

### ✌️ Security

I'd like RW to encourage security throughout -- but *especially* in functions. It's done a tremendous god in getting authentication going and adopted, but I do not believe that many people coming into the serverless world realize that functions are open endpoints into your app and need to be secured thoughtfully. And I know people want to have [basic user/password database auth](https://community.redwoodjs.com/t/simple-email-password-authentication/945), but until `graphql` (which is just a function) can be secured, people need to be aware that one can query all users and their passwords if one does not take care (or worse).

To that end, I could see:

* generators `function -auth - secure` that inject `requireAuth` or a hook into some wrapped JWT functionality that will secure things like functions to process webhooks etc.

I had to write a [bunch of code](https://github.com/dthyresson/deployed/blob/master/api/src/functions/deploy.js) to check a signed JWT and then check that the data in the JWT was valid and belonged to the sub in the JWT etc.

* Adding `requireAuth` to services also maybe as a generator option

As a practice, I'd like to see RW encourage some security best practices - and if they are easy or out-of-the-box, then all the better for adoption.

This also includes introducing HTTP status codes 401, 403, 400 and 500 when appropriate.

###  🔍  SEO

For people new to react, they sometimes don't know that their title, metadata, etc isn't there. 

So you have to use both [React Helmet](https://github.com/nfl/react-helmet) and a prerendering service like [Netlify's](https://docs.netlify.com/site-deploys/post-processing/prerendering/) to get that set.

being able to set these easily in layouts and pages abstracting away Helmet would be great -- as would helpers for basic social media opengraph (Twitter cards, etc).

### 🗂  Prerendering

Tom's already started the [discussion](https://community.redwoodjs.com/t/prerender-proposal/849).

### 📝  Logging

I love Netlify, but it's function logging is meh. And after I looked into using their OAuth API, i saw they stash logs in Firebase and can be pulled in via the deploy API endpoint.

In fact my slow-paced RW project is to have a Netlify build plugin that sends build success/fails which informed reasons to a central space.

Functions leave even more to be desired.

[Winston](https://github.com/winstonjs/winston) looks great.

I know I'd love to see each package have its own logger that I can turn on/off to specify different levels.

And then have my own "app" logger.

I'd be super happy with package can and a timestamp even.

### 🖼  Image Manipulation/Responsive Image Support

I'm not necessarily proposing a full-featured plugin like that of [Gatsby](https://www.gatsbyjs.org/packages/gatsby-plugin-sharp/), but I think some OOB image help would be very useful.

I'd be happy with some sort of general `ImageProvider` similar to auth that wraps services like (Imgix)[https://www.imgix.com/] (which I really like) or [Cloudinary](https://cloudinary.com/features/responsive_images) to render srcsets for responsiveness or other manipulation.

### 📈 Analytics Hooks

Love some easy hooks to send page-level or other events to analytics brokers, like [Segment](https://segment.com/). especially w/in functions and services + easily combined w/ currentUser.

### 🕵️‍♀️  Search hooks

Ok, now rethinking this one. But ... maybe better as build process plugins.

But almost ever app has/wants search and some point.

### ⚖️  Environment Parity (dev/integ/prod)

Big on DX.

This ties into 

* log levels per environment
* what ENVS are used in dev vs Prod
* what database is in dev vs prod

My friend Dan wrote this [parity gem](https://github.com/croaky/parity) for Rails/Heroku and it was great to restore and more around db's.

### ⚙️  3rd Party API Integration Helpers

I keep adding [`node-fetch`](https://github.com/dthyresson/learning-redwood-js/blob/master/api/src/services/weather/weather.js) or `got` or ['graphql-request'](https://github.com/dthyresson/xever-app/blob/master/api/src/lib/hasuraClient.js) when I want to consume data from another source.

Might be nice if something "was just there" that let me make my request. If it even build SDL and a service so I could interct with it via graphql, even better.

###  🪐  The "Other Side"

Love to see an example of another "side" implementation.

For me, the one I think I would do first is a `job`.

I might take in a webhook via a function, send it to a job that calls a service,

Or, can a deploy just the job and api sides in [Render](https://render.com/) and run a script?

That would be great. If it worked w/ AWS SQS and SNS stuff coolio, but just the separate side would be interesting to see.


## 1.0+

* 🎫  Multiple datasources (ie, Fauna + Postgres)

Haven't thought much about tis scenario but think it will be coming sooner vs later.


🥳  Thanks for reading all this!

dthyresson:
Prisma has a nice [page introducing their Public Roadmap] that they maintain using Notion: https://www.notion.so/prismaio/Prisma-public-roadmap-50766227b779464ab98899accb98295f

PR #248, opened 2020-07-14
Improve `body` Function param example to match netlify's API

biphobe:
Running JSON examples after deploying them returns an error:
`json: cannot unmarshal object into Go struct field .body of type string`.

Netlify's Functions, just like AWS Lambda's, need the body to to be a string.


cannikin:
Thanks!

PR #249, opened 2020-07-14
Serverless functions: `statusCode`

Tobbe:
The callback expects `statusCode`, not `status`.
The cookbook recipe has it right https://redwoodjs.com/cookbook/custom-function


Tobbe:
Probably wouldn't be a bad idea to change that code snippet to use `async`, but that's for another PR...

cannikin:
Good catch, thanks!

PR #266, opened 2020-07-30
Fix Primsa Migrate typo

dangdennis:


jtoar:
@dangdennis Nice catch!

PR #268, opened 2020-07-31
tutorials - updated imports of form helps from web -> forms

paulkarayan:
now, without my linter trying to make everything prettier!

@jtoar - i dont know thoughts on trailing commas such as https://github.com/redwoodjs/redwoodjs.com/compare/main...paulkarayan:pk/no-lint?expand=1#diff-81a7b83dbfacb9033bc69bd8730e1c46L1495


thedavidprice:
@paulkarayan thanks for helping with this transition!

Question: are the generators now importing the forms directly or still via @redwoodjs/web? My understanding is that it's currently the latter. But, you know, I'm not quite sure ;)

cannikin:
I saw a lot of trailing commas get removed...I thought our Linter and/or Prettier was putting those in there?

Unless that isn't coming from Redwood's config and it's something in the default config...

paulkarayan:
@cannikin - mine removes them. i didn't know whether to leave them in or not. if there's a standard, let me know. doesn't matter to me either way.

@thedavidprice in the tutorial for forms, they're added by the user. not sure where there are generators that would add this - or if i am misunderstanding.

cannikin:
@thedavidprice Does your Prettier/ESLint add trailing commas, or is it just me?

thedavidprice:
@cannikin Yes, correct. So the commas currently being removed her in `TUTORIAL.md` should be added back. And/or VS Code will just add 'em at time of save.

@paulkarayan Ah, 100% makes sense. So a couple of things then:
- this is definitely where Peter intends to take things, i.e. standalone Forms package and import
- currently, he added support for Forms package to support import via Web package
- the Scaffold generators (e.g. CRUD UI) rely on Forms -- currently importing via Web and will need to be updated sooner than later (could make for a good PR/First issue)

⚠️ --> In the case of writing code that imports Forms directly without Web package (e.g. how you're handling it here, Paul), **currently we do not import the Forms package and it is _not_ included in the Create Redwood App (CRWA) template at the time of install.** So a couple ways forward with your PR here:
- add instruction in the tutorial to `yarn workspace web add @redwoodjs/auth`
- add the package to CRWA repo (I think this is the way to go as a default, but we should check with @peterp to make sure that's his intention)

^^ @cannikin your thoughts about this as well?

thedavidprice:
@peterp Could you help us out on this one? @cannikin and I just discussed as well as tested a fresh install to confirm (for ourselves) the /forms package is being installed with /web package.

From Rob:
> We don’t think we know what’s actually intended by you to do here yet. People may be a little confused if they keep getting warnings in the console to use @redwoodjs/forms instead though. And they probably expect the tutorial to be the latest and greatest recommendations.

So, per your plans:
- is this PR needed and the Tutorial should be updated
- or will /forms always be included in /web, therefore this PR isn't needed?

cannikin:
Sorry @paulkarayan we got our wires crossed there, this PR looks good to me, other than those trailing commas! It looks like those are supposed to be there per Redwood's eslint/prettier config. If you can get those back we're good to go.

peterp:
@thedavidprice @cannikin Yup, `@redwoodjs/forms` is installed via `@redwoodjs/web`, the idea is that if you don't use them that they don't get included in your bundle. Since tree shaking is a mystical thing I would recommend that after a few releases we move the "forms" exports from `@redwoodjs/web` and add them by default into the `./web/package.json`.

Best of both worlds: If you don't use them they're not included, if you do they're available.

TUTORIAL.md
```diff
@@ -1878,9 +1878,9 @@ import {
   TextAreaField,
   Submit,
   FieldError,
-  Label,
-  useMutation
-} from '@redwoodjs/web'
+  Label
+} from '@redwoodjs/forms'
+import useMutation from '@redwoodjs/web'
```
peterp, line 1883:
```suggestion
import { useMutation } from '@redwoodjs/web'
```

TUTORIAL.md
```diff
@@ -2078,10 +2078,9 @@ import {
   Submit,
   FieldError,
   Label,
-  useMutation,
-  FormError,
-} from '@redwoodjs/web'
-
+  FormError
+} from '@redwoodjs/forms'
+import useMutation from '@redwoodjs/web'
```
peterp, line 2083:
```suggestion
import { useMutation } from '@redwoodjs/web'
```

TUTORIAL.md
```diff
@@ -2176,9 +2175,9 @@ import {
   Submit,
   FieldError,
   Label,
-  FormError,
-  useMutation,
-} from '@redwoodjs/web'
+  FormError
+} from '@redwoodjs/forms'
+import useMutation from '@redwoodjs/web'
```
peterp, line 2180:
```suggestion
import { useMutation } from '@redwoodjs/web'
```

PR #273, opened 2020-08-02
g page route param docs

### LINKED ISSUE #260: Update `g page` docs to include {param}
The additions made by https://github.com/redwoodjs/redwood/pull/882 needs to be documented in the cli docs

Tobbe:
Closes #260 

Leaving this as a draft until https://github.com/redwoodjs/redwood/pull/882 is merged

### DIFF
--- a/docs/cliCommands.md
+++ b/docs/cliCommands.md
@@ -653,7 +653,13 @@ Generates a page component and updates the routes.
 yarn rw g page <name> [path]
 ```
 
-If `path` isn't specified, it will be the same as the name. This also updates `Routes.js` in `./web/src`.
+`path` can include a route parameter which will be passed to the generated
+page. The syntax for that is `/path/to/page/{routeParam}/more/path`. You can
+also specify the type of the route parameter if needed: `{routeParam:Int}`. If
+`path` isn't specified, or if it's just a route parameter, it will be derived
+from `name` and the route parameter, if specified, will be added to the end.
+
+This also updates `Routes.js` in `./web/src`.
 
 | Arguments & Options | Description                              |
 | :------------------ | :--------------------------------------- |
@@ -667,7 +673,7 @@ If `path` isn't specified, it will be the same as the name. This also updates `R
 yarn rw d page <name> [path]
 ```
 
-**Example**
+**Examples**
 
 Generating a home page:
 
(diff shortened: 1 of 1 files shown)


docs/cliCommands.md
```diff
@@ -642,7 +642,13 @@ Generates a page component and updates the routes.
 yarn rw g page <name> [path]
 ```
 
-If `path` isn't specified, it will be the same as the name. This also updates `Routes.js` in `./web/src`.
+`path` can include a route parameter which will be passed to the generated
+page. The syntax for that is `/path/to/page/{routeParam}/more/path`. You can
+also specify the type of the route parameter if needed; `{routeParam:Int}`. If
```
jtoar, line 647: Fix typo: semicolon should be colon
```suggestion
also specify the type of the route parameter if needed: `{routeParam:Int}`. If
```

docs/cliCommands.md
```diff
@@ -642,7 +642,13 @@ Generates a page component and updates the routes.
...
+`path` isn't specified, it will be the same as the name. If `path` is only a
```
jtoar, line 648: For consistency (we refer to name as `name` without the article at the end)
```suggestion
`path` isn't specified, it will be the same as `name`. If `path` is only a
```

docs/cliCommands.md
```diff
@@ -703,6 +709,59 @@ const Routes = () => {
...
+$ /redwood-app/node_modules/.bin/rw g page quote {id}
+  ✔ Generating page files...
+    ✔ Writing `./web/src/pages/QuotePage/QuotePage.stories.js`...
+    ✔ Writing `./web/src/pages/QuotePage/QuotePage.test.js`...
+    ✔ Writing `./web/src/pages/QuotePage/QuotePage.js`...
+  ✔ Updating routes file...
+Done in 1.02s.
+```
+
+The page returns jsx telling you where to find it:
```
jtoar, line 726: In this example, you're trying to show the reader that the route parameter is now available, so I'd change this sentence to highlight that difference.

docs/cliCommands.md
```diff
@@ -703,6 +709,59 @@ const Routes = () => {
...
+
+```javascript
```
jtoar, line 728: Highlight the lines of interest for readabillity
```suggestion
```javascript{5,12,14}
```

docs/cliCommands.md
Tobbe: I didn't change this line, but now that you commented on it, it got me thinking... It's not technically 100% correct that the path will be the same as the name. If `name` is `UserDetails`, then the path will be `user-details`.

Is this something we should care about here? Or is it fine to just leave it as is? (or, rather, as you suggested)

docs/cliCommands.md
Tobbe: I changed it to "The generated page will get the route parameter as a prop:". Better?

jtoar: Perfect!

docs/cliCommands.md
jtoar: Ah my bad! That's true&mdash;we could say "derived from name" here and get more specific in another PR?

docs/cliCommands.md
```diff
@@ -653,7 +653,13 @@ Generates a page component and updates the routes.
 yarn rw g page <name> [path]
 ```
 
-If `path` isn't specified, it will be the same as the name. This also updates `Routes.js` in `./web/src`.
+`path` can include a route parameter which will be passed to the generated
+page. The syntax for that is `/path/to/page/{routeParam}/more/path`. You can
+also specify the type of the route parameter if needed: `{routeParam:Int}`. If
+`path` isn't specified, it will be the same as `name`. If `path` is only a
```
jtoar, line 659: @Tobbe If you like my suggestion from https://github.com/redwoodjs/redwoodjs.com/pull/273#discussion_r464549331 
```suggestion
`path` isn't specified, it will be derived from `name`. If `path` is only a
```

docs/cliCommands.md
Tobbe: 👍

docs/cliCommands.md
Tobbe: I like "derived from". The next sentence in the paragraph has the same problem, with using `name` as if it's used exactly as written by the user. And that last sentence was much more difficult to fix. So I rewrote both sentences.

jtoar: @Tobbe Thank you! :rocket:

Tobbe:
@jtoar So, happy with how this looks now?

PR #274, opened 2020-08-02
cliCommands.md: prettier.js format

Tobbe:


jtoar:
@Tobbe now that's some TLC.

PR #275, opened 2020-08-02
Updated contact form section (deprecation notice)

JamesHighsmith:
Got the deprecation notice and wanted to update the docs. 
Deprecation notice, forms have moved:
    Pleas use:
    'import { Form } from "@redwoodjs/forms"'
    instead of:
    'import { Form } from "@redwoodjs/web"'


jtoar:
@JamesHighsmith Thanks a ton!

PR #276, opened 2020-08-02
Update "No API" recipe to use newer build command

### LINKED ISSUE #135: Update "No API" recipe
Once https://github.com/redwoodjs/redwood/pull/576 goes live in a release

jtoar:
Closes #135. We no longer have to specify sides in `yarn rw build`&mdash;Redwood just builds what's there

### DIFF
--- a/cookbook/No_API.md
+++ b/cookbook/No_API.md
@@ -26,7 +26,7 @@ Open up `netlify.toml`. We're going to comment out one line:
 
 ```toml{4}
 [build]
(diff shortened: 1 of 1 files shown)


cannikin:
Hmm...I'd still have them remove the `db up` and `dataMigrate up` parts of the build command...maybe we could add a blockquote to the effect of:

> Redwood is smart and will skip the database migration steps automatically if the `/api` directory doesn't exist, but we like to keep things neat and tidy!

What do you think?

jtoar:
@cannikin Ah right we don't need those commands because there's no database. And I like the blockquote, updating shortly.

PR #279, opened 2020-08-05
Remove fixed position class as it was causing issues with certain layouts

dfundingsland:
This pull request is to address an issue with the menu bar layout. For certain screen sizes - namely iPad or similar sized screens in portrait orientation - the menu bar would resize when toggling the menu. Removing the toggle between fixed and relative classes when clicking on the menu appears to resolve the issue.


cannikin:
Thanks!

Issue #253, opened 2020-07-20
add `yarn rw diagnostics` command to CLI docs


thedavidprice:
See https://github.com/redwoodjs/redwood/pull/842

And this comment: 
https://github.com/redwoodjs/redwood/pull/842#issuecomment-658381749

PR #280, opened 2020-08-05
Add rw diagnostics command to cli docs

forresthayes:
Addresses GH-253

Does this command need a warning? I don't think it's error-prone.

### DIFF
--- a/docs/cliCommands.md
+++ b/docs/cliCommands.md
@@ -352,7 +352,7 @@ yarn rw diagnostics
 
 **Example**
 
-```plaintext{10}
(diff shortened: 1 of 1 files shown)


jtoar:
Nice @forresthayes! Thanks for getting this going. This looks good to go as is, but were you interested in adding an example of running the command, including some example output?

forresthayes:
Good idea! An example would be best. Will reproducing Aldo's Routes.js example suffice?

jtoar:
@forresthayes Yeah that'd be great :+1:

PR #286, opened 2020-08-11
Implements async/await with error handling

alvincrespo:
This PR updates the example tutorial code for the `onSubmit` handler, removing confusion around GraphQL errors for newcomers.

I ran into the following error during the tutorial:

Screen Shot 2020-08-10 at 8 28 02 PM

We need to handle the thrown error by wrapping it in an async/await and catching the error as part of the `onSubmit`.

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -2201,7 +2201,7 @@ const ContactPage = () => {
     try {
       await create({ variables: { input: data } })
     } catch (error) {
-      debugger
+      console.log(error)
     }
   }
 
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -2200,6 +2200,7 @@ const ContactPage = () => {
   const onSubmit = async (data) => {
     try {
       await create({ variables: { input: data } })
+      console.log(data)
     } catch (error) {
       console.log(error)
     }
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -1943,9 +1943,13 @@ The `useMutation` hook returns a couple more elements along with the function to
 const ContactPage = () => {
   const [create, { loading, error }] = useMutation(CREATE_CONTACT)
 
-  const onSubmit = (data) => {
-    create({ variables: { input: data } })
-    console.log(data)
+  const onSubmit = async (data) => {
+    try {
+      await create({ variables: { input: data } })
+      console.log(data)
+    catch (error) {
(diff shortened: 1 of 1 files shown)


thedavidprice:
Hi @alvincrespo! Thank you for this. And, to be clear, is the impetus for this specific to _improved error handling_. My experience with the code itself is that it's working correctly but it effectively blows up if there's an error, correct?

Looping in @cannikin for review.

alvincrespo:
Yeah - basically this blew up on me when moving the email validation from web to api. I thought `useMutation` wouldn't blow up that way, since we're using `error` - but I guess we still need to catch it and handle it. 

Unless I'm doing something wrong, this might be clearer if other newcomers experience the same.

cannikin:
Hmmmm I'm hesitant to add this...I've been leaving error handling out of the tutorial to keep the actual code snippets as succinct as possible. In a real application yes, you should have all of these checks, but we're trying to get people through the major features of the framework as expediently as we can.

What about a blockquote nearby mentioning that ideally this networking code should be made safe against errors, with an example snippet like in your code change?

thedavidprice:
> What about a blockquote nearby mentioning that ideally this networking code should be made safe against errors, with an example snippet like in your code change?

This makes a lot of sense to me, RC. +1

How about on your end @alvincrespo?

alvincrespo:
> we're trying to get people through the major features of the framework as expediently as we can.

That totally makes sense! 

I personally got stuck on the part where I expected the error from the API to show because that GraphQL error page would show up - unlike the video tutorial where you can see the error. So I was trying to find a way to remove any confusion. Making the code change could lead to more confusion as well.

I'll add a blockquote on the page to address this issue. Any recommendations for where it would make the most sense?

cannikin:
> basically this blew up on me when moving the email validation from web to api

So in this section: https://redwoodjs.com/tutorial/saving-data#displaying-server-errors you added the manual handling of the `error` variable but didn't see the message as shown in the screenshot?

image

(Later we convert that error output to use `<FormError>` instead.)

alvincrespo:
Correct. I had that GraphQL page takeover happen. So I needed to add in the error handling to move to the next step of the tutorial.

The odd thing here is that I went back to that step just now and I'm not getting that GraphQL error page takeover anymore. 🤔

alvincrespo:
Closing this one in favor of #290

PR #292, opened 2020-08-14
Typo - also We're FOSS!

rathboma:
Hey redwood team,

Thanks so much for recommending Beekeeper Studio in your docs.
Just fixing a capitalization typo (K vs k). Is it ok if it mentions we're cross-platform and open source?


cannikin:
Absolutely, thanks for correction! :)

PR #305, opened 2020-08-20
Documents Auth Provider Specific Integration incl. Netlify Identity 

dthyresson:
This PR 

* adds examples to setup specific Auth Provider API integration for RBAC support.
* renames the generic "API Specific Integration" to "Auth Provider Specific Integration"
* breaks out the section of protecting functions and services from the Auth0 specific section

Due to an issue with named anchors rendered by Markdown not setting "turbolinks to false" as in:

```
<a href="#auth-provider-specific-integration" data-turbolinks="false">Auth Provider Specific Integration</a>
```

The names anchors for the `Auth Provider Specific Integration` use canonical production env urls which will note exist until deploy.

The anchors otherwise pop to the document head ... as is the case with other named headings I believe.

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -59,6 +59,10 @@ ReactDOM.render(
 )
 ```
 
+#### Netlify Identity Auth Provider Specific Setup
+
+See the Netlify Identity information within this doc's [Auth Provider Specific Integration](https://redwoodjs.com/docs/authentication.html#auth-provider-specific-integration) section.
+
 +++
 
 ### GoTrue-JS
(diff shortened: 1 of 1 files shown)


docs/authentication.md
```diff
@@ -597,7 +600,80 @@ export const getCurrentUser = async (decoded, { type, token }) => {
...
+For example:
+
+```js
+// api/src/lib/auth.js`
+export const getCurrentUser = async (decoded) => {
+  return context.currentUser || { ...decoded, roles: parseJWT({ decoded }).roles }
+}
+```
+
+Now your `currentUser.roles` info will be availble to both `requireAuth()` on teh api side and `hasRole()` on the web side.
```
peterp, line 674:
```suggestion
Now your `currentUser.roles` info will be available to both `requireAuth()` on the api side and `hasRole()` on the web side.
```

dthyresson: Thanks. Fixed typos and squashed commits.

thedavidprice:
This looks great! Huge thanks for making this better, DT!

Merging now.

PR #311, opened 2020-08-21
add schema relations doc

thedavidprice:
used copy from here:
https://community.redwoodjs.com/t/prisma-beta-2-and-rwjs-limited-generator-support-for-relations-with-workarounds/361


docs/schemaRelations.md
```diff
@@ -0,0 +1,145 @@
...
+    id: Int!
+    title: String!
+    body: String!
+    user: User!
+    userId: Int!
+  }
+
+```
+
+The issue is with Redwood’s use of `userID`. We are unable to create a new record by using the foreign key of another table. In this case, where `Post` has a `userId` column, we cannot just set the `userId` and save the record.
```
jtoar, line 91:
```suggestion
The issue is with Redwood’s use of `userId`. We are unable to create a new record by using the foreign key of another table. In this case, where `Post` has a `userId` column, we cannot just set the `userId` and save the record.
```

thedavidprice:
Thanks @jtoar! Merging now.

PR #320, opened 2020-08-22
Roadmap to v1.0

mojombo:
@jtoar Pushing this new branch since yours (#245) is on your fork. I've created GitHub projects for the various areas and linked to those instead of listing issues in this document. Pared down the list to get this out the door, we can do another pass once this is shipped to add things we missed.

### DIFF
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -1,0 +1,234 @@
+# Roadmap to 1.0
+
+The speed we're moving at is pretty remarkable. We want to hit `1.0` by the end of the year. And we think we can do it. But there are a lot of moving parts, each at a different level of maturity. 
+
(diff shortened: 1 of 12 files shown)


thedavidprice:
In general, once this is live I'd like us to make it very prominent: visually highlight, include in Contributing info, etc. To the end, thoughts about putting this at the top of the left nav menu?

Screen Shot 2020-08-23 at 10 21 41 AM

jtoar:
@mojombo The "Last updated 24 Aug 2020" is a nice touch--I'll be sure to check that gets updated when we update this doc.

There were a few edits I made by committing directly to this branch. Is it ok that I did it that way? Otherwise, everything looks good to go to me!

PR #324, opened 2020-08-24
Allow local build without Algolia set up

mojombo:
I found it annoying to have to manually edit code to build the website locally, so I changed the `search.js` file to be able to noop the search stuff if it can't find the API env vars. It's a little gross, but less gross than manually editing code and having to remember not to include those changes in PRs.


peterp:
Closes #181

PR #332, opened 2020-08-27
Updates auth docs for new hasRole support for multiple roles

dthyresson:
See: PR https://github.com/redwoodjs/redwood/pull/1016

When that is merged and released, then these docs update the Auth sections for `hasRole` accordingly.

See PR for details.


dthyresson:
FYI. PR #1016 had been merged so ready when 0.17 is released.

PR #337, opened 2020-09-03
fix typescript docs nav link

forresthayes:
I noticed the TypeScript docs link in the RedwoodJS forum announcement for v0.17.0 is broken, but should be correct.

The TypeScript doc is currently live at https://redwoodjs.com/docs/type-script

I think the dash in "type-script" is a regex quirk.

For reference: [param-case docs](https://github.com/blakeembrey/change-case#options) and the [default split](https://github.com/blakeembrey/change-case/blob/d945e02b515cb8f7797328bcc5d6a29eb6c5832d/packages/no-case/src/index.ts#L11)

### DIFF
--- a/lib/docutron.js
+++ b/lib/docutron.js
@@ -91,7 +91,7 @@ const splitToPages = (markdown, book, options = {}) => {
       title = options.title
     }
 
-    return { href: `/${book}/${paramCase(title)}.html`, title, text: section }
+    return { href: `/${book}/${paramCase(title, {splitRegexp: []})}.html`, title, text: section }
   })
 
   return groups


thedavidprice:
Ah, huge thanks for this but I already snuck in a redirect. Closing this one out. But don't let that discourage you for next time, k? 

😆

forresthayes:
No problem! I had fun digging into Rob's docutron.js

cannikin:
> @forresthayes No problem! I had fun digging into Rob's docutron.js

I'm deeply sorry. 😅

cannikin:
Sorry this took so long to see! So you think this might actually fix #374? What does passing the `splitRegex` option do that `paramCase` doesn't do on its own?

forresthayes:
I think the paramCase function is splitting camel case words by default before replacing spaces with dashes. The splitRegex option lets us override that behavior in favor of preserving a name like TypeScript as one word.

Another example would be the Docs page for [Mocking GraphQL requests](https://redwoodjs.com/docs/mocking-graph-ql-requests).

I think the url for that page should be `redwoodjs.com/docs/mocking-graphql-requests`, but, given the default conversion from camel case, the actual, current address is `https://redwoodjs.com/docs/mocking-graph-ql-requests`

The search results for GraphQL, like TypeScript, are also broken links for me.

PR #338, opened 2020-09-03
fix same page links in docs

forresthayes:
Links to other sections of the same page go to the top of the page instead of the intended target.

For example, in the [Installation](https://redwoodjs.com/docs/redwood-router#installation) section of the Redwood Router doc there is a broken link to the page's final section, "Installation and use outside of a Redwood app".

@jtoar resolved this issue for links in the "on this page" nav with #175

This PR similarly disables Turbolinks for same-page links in the main content.

### DIFF
--- a/lib/docutron.js
+++ b/lib/docutron.js
@@ -8,7 +8,6 @@ const { titleCase } = require('title-case')
 const highlightLines = require('markdown-it-highlight-lines')
 const markdownItTocAndAnchor = require('markdown-it-toc-and-anchor').default
 const markdownItCollapsible = require('markdown-it-collapsible')
-const iterator = require('markdown-it-for-inline')
 const hljs = require('highlight.js')
 const md = require('markdown-it')({
   html: true,
@@ -29,11 +28,6 @@ md.use(markdownItTocAndAnchor, {
   anchorLink: true,
 })
 md.use(markdownItCollapsible)
-md.use(iterator, 'disable-turbolinks-for-same-page-links', 'link_open', (tokens, idx) => {
-  if (tokens[idx].attrGet('href').startsWith('#')) {
-    tokens[idx].attrPush(['data-turbolinks', false])
-  }
-})
 
 const HTML_ROOT = path.join('code', 'html')
 const TEMPLATE_ROOT = path.join('lib', 'templates')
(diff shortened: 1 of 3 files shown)


dthyresson:
This is great! Will help so much by being able to link to headings now.

dthyresson:
I tried

https://deploy-preview-338--redwoodjs.netlify.app/docs/authentication#role-based-access-control-rbac

but this does not seem to go to the target.

forresthayes:
Right! I noticed the section headings have id attributes for targeting _and_ are wrapped in links to themselves (?). I wasn't sure why those links were there. I thought that might be a separate issue.

dthyresson:
> Right! I noticed the section headings have id attributes for targeting _and_ are wrapped in links to themselves (?). I wasn't sure why those links were there. I thought that might be a separate issue.

Yes, shouldn't these be `a name` and not a `href`? And the href wraps a single "#". Hm.

I guess the `a href is ok` ... if there was also an `<a name` inside that `h3`.

```
<h3 id="role-protection-on-functions-services-and-web"><a class="markdownIt-Anchor" href="#role-protection-on-functions-services-and-web">#</a> Role Protection on Functions, Services and Web</h3>
```

Realized that the odd behavior of this heading was due to the collapsed section above not being closed.

cannikin:
The title + link, that's your standard behavior where clicking a title references itself, so it updates the URL with the ID of that element so you can copy+paste that URL to someone else. It's the same thing GitHub does in any markdown that it parses: https://github.com/redwoodjs/redwood#technologies

I don't know that I've ever seen special behavior when you add a `name` attribute to an `<a>`, is that a thing?

forresthayes:
Oh, that makes sense. Would you rather Turbolinks be disabled on those links too?

For reference, [MDN describes the anchor element name attribute as obsolete since HTML5](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attr-name:~:text=nameObsolete%20since%20HTML5)

cannikin:
Turbolinks is usually pretty smart, I'm surprised this isn't working...looks like it *was* an issue in 2018 but should have been fixed with this: https://github.com/turbolinks/turbolinks/pull/348

But yeah it looks like adding a `data-turbolinks="false"` fixes it. Ideally that attribute would only get added if the link is within the same document—not using turbolinks when linking to other pages defeats the whole purpose!

forresthayes:
This alternative approach also disables Turbolinks for the header anchors generated by markdown-it. Anyway, just an option. It does seem like an issue that Turbolinks would handle automatically.

cannikin:
Thanks @forresthayes!

PR #342, opened 2020-09-06
Fix typo in useAuth() hook

clairefro:
BEFORE 
`userAuth()` 
AFTER 
`useAuth()`

`useAuth` appears to be the correct hook name as referenced elsewhere

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -711,7 +711,7 @@ You can also protect routes:
 </Router>
 ```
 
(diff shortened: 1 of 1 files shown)


cannikin:
Eagle-eye @clairefro! Thanks for the fix! :)

dthyresson:
Thanks @clairefro! I have updated this in my RBAC cookbook branch as well.

PR #343, opened 2020-09-06
fix grammar in 'Saving Data' tutorial section

forresthayes:
This only changes the last sentence in the paragraph.

> We felt that only having one update input, while maybe not pedantically the absolute **correct** way to create a GraphQL API, ~~we felt it~~ was a good compromise for ~~an~~ optimal developer experience.


cannikin:
Yeah this reads better, thanks!

PR #351, opened 2020-09-09
rename diagnostics to check

forresthayes:
docs update for issue redwoodjs/redwood#957, and PR redwoodjs/redwood#1112


dthyresson:
Will this pr have to wait to be merged until the actual CLI command changes from `diagnostics` to `check`?

https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/diagnostics.js

forresthayes:
I think so. I made a [PR](redwoodjs/redwood#1112)  to rename the command in the redwood repo, so this will be ready to go if that is approved.

I can mark this as a draft for the meantime, sorry.

PR #355, opened 2020-09-10
typo in schemaRelations.md

qooqu:
In the workaround code, const is commented out.

(First pull request ever ... apologies if I screwed something up)


thedavidprice:
@qooqu Great catch and this PR is fantastic 🚀

We get especially excited when people tell us that Redwood is their first-ever open source contribution! Welcome!! And look forward to more 😆

PR #359, opened 2020-09-11
add style object option to Flash doc

forresthayes:
Brings the docs up to date with [this PR](https://github.com/redwoodjs/redwood/pull/1133), adding the Flash inline-style option, slated for the next release.


jtoar:
@forresthayes Thanks!

PR #361, opened 2020-09-12
Add db import for services section

ryanmdoyle:
Added the db import and updated description to say no "other" imports were omitted.

Resolves redwoodjs/redwood#812

### DIFF
--- a/code/html/index.html
+++ b/code/html/index.html
@@ -272,8 +272,6 @@ export const Success = ({ users }) => {
     </div>
     <div class="md:w-3/5 md:pl-4 mt-4 md:mt-0">
       <pre><code class="javascript" data-target="application.code">// api/src/graphql/users.sdl.js
-import { db } from 'src/lib/db'
-
 export const schema = gql`
   type User {
     id: Int!
@@ -285,6 +283,8 @@ export const schema = gql`
 `
 
 // api/src/services/users/users.js
+import { db } from 'src/lib/db'
+
 export const users = () => {
   return db.user.findMany()
 }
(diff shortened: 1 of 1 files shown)


dthyresson:
@ryanmdoyle 

Current screenshot and copy:

Screen Shot 2020-09-12 at 13 34 35

```js
    <div class="md:w-3/5 md:pl-4 mt-4 md:mt-0">
      <pre><code class="javascript" data-target="application.code">// api/src/graphql/users.sdl.js
import { db } from 'src/lib/db'

export const schema = gql`
  type User {
    id: Int!
    name: String!
  }
  type Query {
    users: [User]
  }
`

// api/src/services/users/users.js
export const users = () => {
  return db.user.findMany()
}

// web/src/components/UsersCell/UsersCell.js
export const QUERY = gql`
  query USERS {
    users {
      id
      name
    }
  }
`
export const Success = ({ users }) => {
  return JSON.stringify(users)
}
```

Even though the code section looks like a single file, it actually represents three and I think that

```
import { db } from 'src/lib/db'
```

is better suited to be associated with `api/src/services/users/users.js` since that is where the `db` import is used and people won't think it is part of the SDL:

```
// api/src/services/users/users.js
import { db } from 'src/lib/db'

export const users = () => {
  return db.user.findMany()
}
``

ryanmdoyle:
Of course. I'll make the change. Thanks for the guidance!

ryanmdoyle:
@dthyresson I updated the import to be under the comment for the actual services file, so I think it's good!

dthyresson (approved):
LGTM, thanks!

Issue #366, opened 2020-09-18
Docs - Deploy: Two or three deploy targets?


kimadeline:
https://redwoodjs.com/docs/deploy:

> Currently, there are two officially supported deploy targets:
>   - Netlify
>   - Vercel
>   - AWS-Serverless

Almost got me there ✌️

PR #367, opened 2020-09-18
Update deploy target count in the docs

kimadeline:
Update the [deploy docs](https://redwoodjs.com/docs/deploy) to address #366.

I would say that this is the smallest PR I've ever created, but I don't think so 😆


peterp:
lol, my bad! Thanks for this :)

PR #373, opened 2020-09-18
fix page title formatting for stand-alone pages like the Roadmap

forresthayes:
It's a little thing, but the [Roadmap page](https://redwoodjs.com/roadmap) title has a leading dash in a browser tab, ` - Roadmap : RedwoodJS Docs`, because, unlike most doc pages, it doesn't have a 'book' context.


cannikin:
Great catch!

PR #382, opened 2020-09-28
add aws_serverless to `rw g deploy` provider options in CLI doc

forresthayes:


cannikin:
@thedavidprice is this legit, can people 100% use the generator for serverless deploy now? We should definitely merge this in if that's the case!

forresthayes:
Sorry if it's not ready yet, I was reading the Deploy docs and noticed the option there: https://redwoodjs.com/docs/deploy#aws-serverless-deploy

thedavidprice:
2 legit 2 quit 🚀

Thanks!

PR #385, opened 2020-10-01
Improved documentation and corrected some typo errors

bboygary:


cannikin:
Thanks! Can't believe we've missed some of those things for so long!

PR #387, opened 2020-10-01
Simplify contributions (no need for GitHub token)

Tobbe:
I wanted to make it easier to contribute to the docs by providing instructions on how to get around the need for a GitHub token. (Like we already do for the Algolia token.)

An alternative to the two workarounds in the readme would be to make the code gracefully handle the case where the tokens aren't provided. If that's the way we want to go instead I'm happy to update this PR with the needed changes

Also added a friendly reminder to run `yarn install`

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,8 +10,6 @@ Other documentation is pulled from various READMEs in the main [redwoodjs/redwoo
 
 This codebase is built with https://cameronjs.com and relies on plain HTML pages and Javascript with a couple helpers built in to abstract things like partials and layouts. We use https://stimulusjs.org for the few sprinkles of interactivity throughout the site.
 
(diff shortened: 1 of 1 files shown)


cannikin:
Actually the first half of those instructions (for skipping Algolia search) are obsolete! @mojombo had a PR that skips search automatically if the `ENV` var doesn't exist. We should probably make sure it's commented out in `.env.example` so that someone copying that to `.env` doesn't just get errors because the var is set to some generic placeholder.

We should probably do something similar for this, rather than explain a bunch of code changes on how to skip. Makes it much friendlier for new contributors...

Tobbe:
I'll see if I can find Tom's PR and then do something similar with the Octokit stuff

Tobbe:
At first I thought you meant Tom had an open PR for those changes, but it had already been merged. 
And for GitHub, if you don't provide a token it will just run on the free tier instead, which is perfectly fine for developing locally.

So I just removed that whole section from the readme. Less information overload 🙂

thedavidprice (approved):

PR #388, opened 2020-10-01
No page break headers in code blocks

Tobbe:
When parsing markdown files there is a function that looks for markdown headers and will automatically split the document in several pages at each header.

The problem is when there is a code block (text between triple backticks) that starts with the hash (#) character. This will incorrectly be interpreted as a markdown header and the page will be split in to several pages.

I made a quick edit to one of the existing documents to demonstrate.

This is how it should work (and how it works with my fix)

after_code_header_fix

Notice the `# Terminal comment` string in the code block

Without this PR fix this is what happens:

Page 1
before_code_header_fix_page1

Page 2
before_code_header_fix_page2

As you can see the page is split in two parts, and the formatting is all messed up on the second page

### DIFF
--- a/lib/docutron.js
+++ b/lib/docutron.js
@@ -74,12 +74,18 @@ const splitToPages = (markdown, book, options = {}) => {
     }
   }
 
+  let isCodeBlock = false
+
   markdown.split('\n').forEach((line, index) => {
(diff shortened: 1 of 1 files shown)


cannikin:
Thanks for fixing that!

PR #401, opened 2020-10-06
Add missing closing tag

stav:
Add missing closing tag for some contrived, tutorial-like reason in Route Param Types


stav:
Ah, I guess that the closing tag in not actually "missing" since there could be content inside.  But, I guess with the closing slash it makes it more consistent with the tutorial code.

cannikin:
Good catch, thanks!

PR #405, opened 2020-10-07
Fix tutorial and doc links for local dev

Tobbe:
When running `yarn dev` the two big buttons that link to the tutorial and the docs don't work. The links go to `/tutorial` and `/docs` which don't exist. I guess there is some redirect rule setup somewhere for redwoodjs.com that make them work in prod.

image

I've added a middleware to live-server that does that redirect when running locally.

I'm not super happy with how I had to specify the path to the middleware 😞 Please let me know if you want me to change this, and if so, any hints on how would be very appreciated.

Another thing is that this doesn't update the browser url. It still says localhost:8080/tutorial instead of localhost:8080/tutorial/welcome-to-redwood.html. I am not sure why. I thought 301 and 302 redirects would cause the url to update...

### DIFF
--- a/lib/middleware/use-extension.js
+++ b/lib/middleware/use-extension.js
@@ -1,0 +1,37 @@
+// https://github.com/tapio/live-server/issues/244
+
+const fs = require('fs')
+const path = require('path')
+const rootDirectory = './publish/'
+const extensions = ['html']
+
+module.exports = function (req, res, next) {
+  if (req.method !== 'GET' && req.method !== 'HEAD') {
+    return next()
+  }
+
+  if (req.url !== '/' && path.extname(req.url) === '') {
+    const requestedPath = req.url.replace('/', '')
+    let i = 0
+    const check = () => {
+      const path = rootDirectory + requestedPath + '.' + extensions[i]
+
+      fs.access(path, (err) => {
+        if (!err) {
+          req.url += '.' + extensions[i]
+          next()
+        } else {
+          if (++i >= extensions.length) {
+            next()
+          } else {
+            check()
+          }
+        }
+      })
+    }
+
+    check()
+  } else {
+    next()
+  }
+}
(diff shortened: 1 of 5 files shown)


cannikin:
HMMMM  What if the middleware read in `netlify.toml` and used the `[[redirects]]` that are listed there, rather than just hardcode these two (`/tutorial` and `/docs`)?

cannikin:
> Another thing is that this doesn't update the browser url. It still says localhost:8080/tutorial instead of localhost:8080/tutorial/welcome-to-redwood.html. I am not sure why. I thought 301 and 302 redirects would cause the url to update...

If I remember right, you need to send back an extra header to tell Turbolinks to update the URL. Here's the definition of the `/tutorial -> /tutorial/welcome-to-redwood` redirect from `netlify.toml`:

```toml
[[redirects]]
  from = "/tutorial"
  to = "/tutorial/welcome-to-redwood"
  status = 301
  force = true

[[headers]]
  for = "/tutorial/welcome-to-redwood"
  [headers.values]
    Turbolinks-Location = "/tutorial/welcome-to-redwood"
```

So the URL you're going TO needs to return that header in order to get the URL updated.

Tobbe:
I like the idea of reading from `netlify.toml`. I'll have a stab at it! 
Also found this https://github.com/turbolinks/turbolinks#following-redirects

Tobbe:
The thing I like the least about the current PR solution is the relative paths to the middlewares. 

What I could do is write a wrapper around live-server that more reliably looks those paths up, and passes them on to live-server. Do you think I should do that? Or is the current solution good enough?

cannikin:
> The thing I like the least about the current PR solution is the relative paths to the middlewares.

Ugh yeah those relative paths are gross...but having to create a whole wrapper around it just for the paths...also ugh.

I'm okay with the relative paths if you are!

cannikin (approved):

Tobbe:
> I'm okay with the relative paths if you are!

I am. Please LGTM 🙂 

Can always revisit if needed in the future

cannikin:
MERGED

PR #406, opened 2020-10-08
fix typos on the RBCA cookbook

esteban-url:
fixing some small typos that I noticed


dthyresson (approved):
Thanks for all those corrections -- I am such a terrible typist. ;)

esteban-url:
:) glad to be of help. I'll probably do a few more of these as I properly go through the tutorial and docs.

PR #426, opened 2020-10-22
Add steps to invoke identity functions to RBAC cookbook

Irev-Dev:
I put some information on the forums
https://community.redwoodjs.com/t/testing-netlify-identity-triggers/1337
and @dthyresson asked I add some of the steps to the RBAC docs, I wasn't too sure how much detail to add so please let me know what to tweak.

Obviously my changes only apply to `cookbook/Role-based_Access_Control.md` though a number of files under version control seem to change with `yarn build` I wasn't sure if they should be commited or not, they're all in d7469e6. Let me know if I need to revert this commit.

### DIFF
--- a/cookbook/Role-based_Access_Control.md
+++ b/cookbook/Role-based_Access_Control.md
@@ -634,6 +634,8 @@ netlify functions:invoke <function-name> --port 8910
 ```
 `<function-name>` should be replaced by `identity-validate`, `identity-signup`, `identity-login` or your own function.
 
+Note that the netlify-cli does not generate fake user data for each invocation of an identity function. It always provides the same `Test Person` data.
+
 ## Additional Resources
 
 - [RBAC Example & Demo Site](https://redwoodblog-with-identity.netlify.app/)
--- a/cookbook/Role-based_Access_Control.md
+++ b/cookbook/Role-based_Access_Control.md
@@ -618,6 +618,22 @@ export const handler = async (req, _context) => {
 }
 ```
 
+#### How to invoke serverless functions while in dev
+
+So long as `yarn rw dev` is running, `netlify-cli` can be used to invoke your function. Steps are:
+
+```terminal
+# Install the cli
+yarn add netlify-cli -g
+
+# Rebuild api after any changes to /functions
+yarn rw build api
+
+# Invoke your function with the CLI, pointing it to the rw dev port
+netlify functions:invoke <function-name> --port 8910
+```
+`<function-name>` should be replaced by `identity-validate`, `identity-signup`, `identity-login` or your own function.
+
 ## Additional Resources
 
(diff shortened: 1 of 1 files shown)


Irev-Dev:
https://deploy-preview-426--redwoodjs.netlify.app/cookbook/role-based-access-control-rbac#how-to-invoke-serverless-functions-while-in-dev

Tobbe:
@Irev-Dev Thanks for putting this together!

> I wasn't sure if they should be commited or not, they're all in d7469e6. Let me know if I need to revert this commit.

Yeah, if you could only include `cookbook/Role-based_Access_Control.md` that would be great 🙂 If it's not too much trouble.  Those other files will be updated by our build script.

Irev-Dev:
Thanks @Tobbe, that commit has been reset.

thedavidprice:
Just an fyi that @dthyresson has been heads down on a project this past week. Should be reemerging next week. And I've assigned him as a reviewer as well.

Do ping him again into next week if/as needed.

dthyresson (changes requested):
@Irev-Dev Thanks for this most welcome addition. Being able to test out the Netlify Identity functions is really useful.

Howe3ver, please remind me, is not the content of the `identity` that is used via netlify-cli a mocked up dummy user account?

See: https://github.com/netlify/cli/blob/487d11dcaf37a1035b953ff5cd0ce2612aa7fe47/src/commands/functions/invoke.js#L67

```
if (name === 'identity') {
        // https://www.netlify.com/docs/functions/#identity-event-functions
        body.event = event
        body.user = {
          id: '1111a1a1-a11a-1111-aa11-aaa11111a11a',
          aud: '',
          role: '',
          email: 'foo@trust-this-company.com',
          app_metadata: {
            provider: 'email',
          },
          user_metadata: {
            full_name: 'Test Person',
          },
          created_at: new Date(Date.now()).toISOString(),
          update_at: new Date(Date.now()).toISOString(),
        }
      }
```

So, I think it would be worthwhile to make not of that in case someone wonders where this "Test Person" from "trust-this-company.com".

I am actually really surprised Nertlify did not use something like https://www.npmjs.com/package/faker to create random user accounts since this returns to same one over and over.

Thanks.

Irev-Dev:
@dthyresson I added a note that netlify does not generate data each time
`Note that the netlify-cli does not generate fake user data for each invocation of an identity function. It always provides the same 'Test Person' data.`
And I confirmed that's the case with all three identity functions locally. the user object is always the same.
I considered including the user object in the docs, But this seemed at risk of going stale quickly.

Happy to make more adjustments.

Obviously I can't speak for netlify, but maybe this was the more conservative approach for them since, if invoking these is for testing db writes, Maybe their thoughts are the generated data will make clean up harder for their users, even if it is a staging/test evironment 🤷

Tobbe:
@dthyresson Do you think this is ready to be merged?

PR #436, opened 2020-10-28
Indicate that `browser.open` is true by default

jvanbaarsen:
In the docs it says that the `browser.open` is false by default, but in the actually released version it is `true` by default.

### DIFF
--- a/docs/appConfiguration.md
+++ b/docs/appConfiguration.md
@@ -114,7 +114,7 @@ Configuration for the browser target.
 
 | Key    | Description                                                       | Default | Context       |
 | :----- | :---------------------------------------------------------------- | :------ | :------------ |
-| `open` | Open the browser to web's `host:port` after the dev server starts | `true` | `development` |
+| `open` | Open the browser to web's `host:port` after the dev server starts | `false` | `development` |
 
 ### open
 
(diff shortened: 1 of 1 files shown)


thedavidprice:
Hi @jvanbaarsen Ah, I see the confusion here. Let's see if we can figure out a better way to communicate this.

In the code (for example, if you were to remove the config from redwood.toml all together), the default _is_ false:
https://github.com/redwoodjs/redwood/blob/49c3afecc210709641dd340b974c86251ed207dc/packages/internal/src/config.ts#L59-L60
```ts
  browser: {
    open: false,
```

I think you're referring here to the implemented default setting for a new installation being `true`, correct?

So the change you've suggested is not the case. But can you think of a better way we can communicate all this in the docs?

jvanbaarsen:
@thedavidprice Ah check! That makes sense. I was indeed referring to the way a newly generated app is dealing with this. I'll see if I can come up with a better way of wording this!

docs/appConfiguration.md
```diff
@@ -127,6 +127,8 @@ Setting `open` to `true` like this will open the browser to web's `host:port` (b
 
 You can also provide the name of a browser to use instead of the system default. E.g., `open = 'Firefox'` will open Firefox regardless of which browser's the default on your system.
 
+When you generate a new app the `open` value will be set to `true` by default for you. When you remove the `open` key from the `redwood.toml` file we will default to `false`.
```
thedavidprice, line 130:
```suggestion
When you generate a new app, the `open` value is set to `true`. If you delete the `open` config from `redwood.toml`, it will default to `false`. For example, removing the line `open = true` disables automatic browser opening.
```

docs/appConfiguration.md
```diff
@@ -127,6 +127,8 @@ Setting `open` to `true` like this will open the browser to web's `host:port` (b
 
 You can also provide the name of a browser to use instead of the system default. E.g., `open = 'Firefox'` will open Firefox regardless of which browser's the default on your system.
 
+When you generate a new app, the `open` value is set to `true`. If you delete the `open` config from `redwood.toml`, it will default to `false`. For example, removing the line `open = true` disables automatic browser opening.
```
thedavidprice, line 130:
```suggestion
> When you generate a new app, the `open` value is set to `true`. If you delete the `open` config from `redwood.toml`, it will default to `false`. For example, removing the line `open = true` disables automatic browser opening.
```

thedavidprice:
Nice. I attempted some tweaks (at the risk of getting to specific). Take a look at the code, which is reflective of the screenshot below. y/n/maybe?

Screen Shot 2020-10-29 at 2 05 16 PM

jvanbaarsen:
@thedavidprice Yes perfect :)

PR #438, opened 2020-10-31
Update latest NodeJS LTS version from v12 to v14.

miku86:
NodeJS updated their LTS version to v14.

The tutorial says v12, what could lead to confusion when users go to the NodeJS homepage and find v14 to be the latest LTS version.

Source: https://github.com/nodejs/Release
Source: https://nodejs.org/


Tobbe:
Thanks @miku86!

PR #439, opened 2020-10-31
Add missing `const`

jvanbaarsen:
In one of the example the `const` was missing before a method declaration.


cannikin:
Good catch, thanks!

PR #444, opened 2020-11-05
update link in Cells doc

forresthayes:


Tobbe:
Thanks! I've been meaning to do this, but kept forgetting 🙈

PR #445, opened 2020-11-06
change login anchors to buttons in tutorial

ogdenstudios:
The tutorial uses anchor tags in the authentication section to trigger Netlify identity modals. I think the best practice in this case is to use `<button>` elements instead, and it looks like that's what's in the code sample in `docs/authentication.md` as well.

Hope this is a useful PR, thanks for your consideration!


cannikin:
I probably should have done this to start with! 😅 Thanks!

PR #453, opened 2020-11-10
Tutorial: Fix label name props

Tobbe:
The `<Label>` name props all had the same value


cannikin:
Oh geez how did we miss THAT for so long??

PR #454, opened 2020-11-10
Fix :: Focusing on the search field on any keyboard layout

qWici:
The search input was not focused when "/" key was pressed. I don't know if it's Chrome bug or what, but when I press slash `event.key` was "." (dot). Any another keys was in another language. Looks like I'm typing with wrong layout, but when I trying to type something in search input or Chrome console - layout was English. So I propose to use `code` property of events.

**To Reproduce**

1. Go to home page
2. Press slash

**Expected behavior**
Search input focused

**Environment**
 - OS: Ubuntu 20.04.1 LTS
 - Browser: Version 86.0.4240.111 (Official Build) (64-bit)

### DIFF
--- a/code/javascripts/controllers/application_controller.js
+++ b/code/javascripts/controllers/application_controller.js
@@ -32,7 +32,7 @@ export default class extends Controller {
   }
 
   focusSearch(event) {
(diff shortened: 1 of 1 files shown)


cannikin:
Great catch, thanks!

PR #455, opened 2020-11-12
Tutorial/authentication

brentguf:
Hi, 

And finally some for the Authentication part!

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -2799,7 +2799,9 @@ export default BlogLayout
 
 ![Logged in email](https://user-images.githubusercontent.com/300/82389433-05b2e680-99f1-11ea-9d01-456cad508c80.png)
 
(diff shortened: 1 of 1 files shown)


TUTORIAL.md
```diff
@@ -2534,6 +2536,7 @@ export const requireAuth = () => {
     throw new AuthenticationError("You don't have permission to do that.")
   }
 }
+
```
brentguf, line 2539: I didn't know if I had to touch this part. It's not up-to-date anymore with what's currently generated in `auth.js`, but because it looks quite as a WIP (the docs and the code), I'm not sure if this is still going to be entirely rewritten.

TUTORIAL.md
```diff
@@ -2511,7 +2513,7 @@ yarn rw g auth netlify
 
 The generator adds one file and modifies a couple others.
 
-> Don't see any changes?
+> **Are you on the latest Redwood?**
```
brentguf, line 2516: Not even sure if this is still relevant as we're currently already on 0.20.0.

PR #464, opened 2020-11-20
docs: remove Prisma Studio warning

Jolg42:
Prisma Studio is now stable 💚

https://www.prisma.io/blog/prisma-studio-3rtf78dg99fe


janpio (approved):

thedavidprice:
@Jolg42 🚀 Fancy seeing you here! And thanks so much.

Here's the actual command we're calling — looks current and does not require modification, correct?
https://github.com/redwoodjs/redwood/blob/main/packages/cli/src/commands/dbCommands/studio.js

Jolg42:
Indeed all good 👍

Have a great weekend 😁

PR #473, opened 2020-11-30
Update TUTORIAL.md

AsdethPrime:
Change reference from api/prisma to api/db because the cli currently creates a folder called db rather than prisma. 

Change the migration command line code from 

Yarn redwood save create post

To 

Yarn redwood save “create post” for clarity


dthyresson:
LGTM. Merging.

PR #490, opened 2020-12-13
Update serverlessFunctions.md

noobling:
The sample code didn't seem to work for me. I've updated it to an example that did work.

### DIFF
--- a/docs/serverlessFunctions.md
+++ b/docs/serverlessFunctions.md
@@ -11,11 +11,13 @@ at `http://localhost:8911/graphql`.
 The `./api` directory is watched for modifications, when they are
 detected the modules are reimported.
 
+You can use code in `./api/src` e.g. `import { db } from 'src/lib/db'`
+
 A lambda function must export a `handler`. You can execute the
 supplied callback function to return a response:
 
 ```js
-export const handler = (event, context, callback) => {
-  return callback(null, { statusCode: 200, body: 'Hello, world' })
+export const handler = (event, context) => {
+  return { statusCode: 200, body: 'Hello, world' }
 }
 ```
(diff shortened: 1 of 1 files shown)


Tobbe:
Hi @noobling 
Where did you test your serverless function? Where didn't it work?
Netlify still say they support the callback format (The docs here https://docs.netlify.com/functions/build-with-javascript/#synchronous-function-format links to this forum topic that has the callback syntax as an example https://community.netlify.com/t/support-guide-how-do-i-write-a-javascript-lambda-function/24106)

noobling:
Hey it was when I ran the code locally with rw dev. This was with the latest versions of redwood

dthyresson:
@Tobbe and @noobling I think the change is welcome. 

But, should it not be async perhaps -- or at least suggest it might be, especially since it refers to using the Prisma db client which I believe will need an await?

The generators will create:

```
export const handler = async (event, context) => {
  return {
    statusCode: 200,
    body: JSON.stringify({
      data: '${name} function',
    }),
  }
}
```

Perhaps the docs should be:

```
```
export const handler = async (event, context) => {
  return {
    statusCode: 200,
    body: JSON.stringify({
      data: 'Serverless function',
    }),
  }
}
```

**Comment by noobling** [comment]
Sure updated

File: docs/serverlessFunctions.md [hunk 1]
```diff
@@ -11,11 +11,23 @@ at `http://localhost:8911/graphql`.
...
+    body: JSON.stringify({
+      data: 'Serverless function',
+    }),
+  }
 }
-```
\ No newline at end of file
+```
+
+You can also use the generator to create the function.
```
Tobbe, line 29:
```suggestion
Redwood comes with a generator to help you create serverless functions
```

Tobbe:
Thanks for the PR @noobling

PR #495, opened 2020-12-19
fix broken Cells doc links to withCell source code

forresthayes:
I think `withCell.js` was renamed `withCellHOC.tsx`


dthyresson:
@forresthayes  Thanks for the update/correction!

PR #501, opened 2020-12-29
New Auth - Ethereum

pi0neerpat:
This PR depends on https://github.com/redwoodjs/redwood/pull/1594

Will convert from draft once that PR is finished


thedavidprice:
Thanks again @pi0neerpat 

@Tobbe we don't have a good process, but often we hold back docs that are release dependant and merging when the feature/code is released in new version. All good on this one for now. But suggestions on a process that's easy and makes sense? Maybe a label?

Tobbe:
Ahh, yes, sorry. I knew that 🙁 Too excited to get this out! 😄 

The safest process is automation. When we get around to automating the release process it would be super cool if it could look at new PRs going in to the release and see if they have any linked doc-PRs, and then merge those.

Issue #505, opened 2021-01-07
findOne() -> findUnique()


cannikin:
As of 2.12.0 Prisma is deprecating `findOne()` in favor of `findUnique()`. We've got a couple instances of `findOne()` in several docs.

96RadhikaJadhav:
@cannikin  Can I work on the issue?

thedavidprice:
@96RadhikaJadhav please do! Thank you for offering. I'm assigning to you now.

- do reach out to @cannikin directly for help and PR review, although others might step in as well
- do **not** change the Tutorial — that will first need code changes and then copy updates
- do look through everything in [Docs/](https://github.com/redwoodjs/redwoodjs.com/tree/main/docs)

🚀

thedavidprice:
Ah, sorry @96RadhikaJadhav I was moving quickly and didn't see #506 is already in progress. 😞 

Please feel free to jump on any docs related issues as they come! We're pretty loose with assignments as docs change/move quickly.

I saw another Issue that is also a good first issue. Will `@` you from #486

cannikin:
Sorry I missed this request @96RadhikaJadhav, it came in after I went to bed! But like @thedavidprice said, feel free to jump on anything that looks interesting to you, you don't need to wait for us. 🙂

PR #506, opened 2021-01-08
Replaced findOne with findUnique in docs

adithyasunil26:
Prisma release 2.12.0 renames findOne to findUnique and deprecates findOne. I have simply replaced various instances of findOne in the docs with findUnique.
This should close issue #505 .

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -922,7 +922,7 @@ export const posts = () => {
 }
 
 export const post = ({ id }) => {
-  return db.post.findOne({
+  return db.post.findUnique({
     where: { id },
   })
 }
@@ -2647,7 +2647,7 @@ export const posts = () => {
 }
 
 export const post = ({ id }) => {
-  return db.post.findOne({
+  return db.post.findUnique({
     where: { id },
   })
 }
@@ -2675,7 +2675,7 @@ export const deletePost = ({ id }) => {
 }
 
 export const Post = {
-  user: (_obj, { root }) => db.post.findOne({ where: { id: root.id } }).user(),
+  user: (_obj, { root }) => db.post.findUnique({ where: { id: root.id } }).user(),
 }
 ```
 
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -922,7 +922,7 @@ export const posts = () => {
 }
 
 export const post = ({ id }) => {
-  return db.post.findUnique({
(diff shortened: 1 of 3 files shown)


docs/cliCommands.md
```diff
@@ -1064,7 +1064,7 @@ export const users = () => {
 
 export const User = {
   profile: (_obj, { root }) => {
-    db.user.findOne({ where: { id: root.id } }).profile(),
+    db.user.findUnique({ where: { id: root.id } }).profile(),
```
dthyresson, line 1067: All these updates look good, but @cannikin and @thedavidprice have the RedwoodJS generators been updated to match yet?

I can see: 

https://github.com/redwoodjs/redwood/blob/0288d7a75513e0d95269f46130d212ad49c57f7b/packages/cli/src/commands/generate/service/templates/service.ts.template#L48

that `findOne` is still part of the service template. 

Or can this documentation go live ahead of that?

thedavidprice: Let's keep this one. It won't be accurate in the near-term, but this is on Rob's radar and I expect the generators to be updated soon.

Also, I don't think this is a high-traffic doc or a line of code that would confuse people in a significant way given a discrepency.

thedavidprice: Marking this comment resolved.

TUTORIAL.md
```diff
@@ -2675,7 +2675,7 @@ export const deletePost = ({ id }) => {
 }
 
 export const Post = {
-  user: (_obj, { root }) => db.post.findOne({ where: { id: root.id } }).user(),
+  user: (_obj, { root }) => db.post.findUnique({ where: { id: root.id } }).user(),
```
thedavidprice, line 2678: Ah, the Tutorial.md file should **NOT** be included in this PR. The code will need to be updated first, then the Tutorial can be updated to reflect changes in the code.

So please revert these changes. Thanks! And sorry for the confusion.

thedavidprice (changes requested):
See comment about not changing Tutorial.md in this PR

TUTORIAL.md
adithyasunil26: Okay. Reverted the changes. Always happy to contribute.

thedavidprice (approved):

thedavidprice:
Looks great. Thank you @adithyasunil26! Merging now.

PR #509, opened 2021-01-09
Remove key prop from BlogPost

forresthayes:
I think we can remove the `key` prop from the `BlogPost` component in the `BlogPostsCell` code sample at line 218. 

I'm following along with the repo supplied at the start of Tutorial2, and the `key` prop isn't passed [there](https://github.com/redwoodjs/redwood-tutorial/blob/81372de63b850ed7bd72ec9b3005eb5a0cb5dac8/web/src/components/BlogPostsCell/BlogPostsCell.js#L26).


Tobbe:
You are correct. `key` has moved up to the `<div>`, where it should be now.

PR #512, opened 2021-01-09
Fix consistent use of colons before code blocks

dijonmusters:
Could be entirely intentional but just a couple of inconsistent colons before code blocks 👍


cannikin:
Thank you!

PR #513, opened 2021-01-09
Fix BlogPostsCell Success test

forresthayes:
Change the `posts.forEach` loop to reference the truncatedBody of each post in turn, instead of only the first post (`posts[0]`).


Tobbe:
Good catch! Thanks.

PR #514, opened 2021-01-09
Fix BlogPostsCell test regex

forresthayes:
I'm a regex novice, but I'm not sure the regex in the BlogPostsCell Success test needs to include the 'lazy' switch (`?`) after the star `*` here.


cannikin:
Yeah you're right, it's not necessary here. I always default to adding the non-greedy `?`, it's just become a reflex!

PR #515, opened 2021-01-10
Update TUTORIAL2.md

forresthayes:
To double check the test coverage, we would modify `BlogPost.js`, where the `summary` prop was added, and not the `BlogPostCell`, right?

And later, I think the initial tests of the summary prop cover the `BlogPostsCell` instead of `BlogPostCell`.


forresthayes:
My mistake. It's `BlogPostsCell` in both cases.

Tobbe:
🦅 👁️

PR #517, opened 2021-01-12
Fix foreign key formatting

dijonmusters:
Just a couple of formatting issues I found while going through: https://redwoodjs.com/docs/schema-relations


Tobbe:
Thank you!

PR #524, opened 2021-01-17
Docs: TypeScript docs updates

msutkowski:
- Add notes about minimum required node version for local dev
- Add notes about types for `jest`
- Addresses points from https://github.com/redwoodjs/redwood/issues/1400

There is an additional issue here when working with scaffolded code and the file structure, but I don't know if it is worth addressing with this PR. https://github.com/redwoodjs/redwood/issues/234#issuecomment-620906084 regarding the DirectoryName convention. TS users would just have to:
 1. add an `index.ts` to `src/components/WhateverComponent'`
 2. `export { default as default } from './WhateverComponent'`

As a side note, I didn't see any specific plan for handling this in the tracking issue: https://github.com/redwoodjs/redwood/issues/523

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ Other documentation is pulled from various READMEs in the main [redwoodjs/redwoo
 
 This codebase is built with https://cameronjs.com and relies on plain HTML pages and Javascript with a couple helpers built in to abstract things like partials and layouts. We use https://stimulusjs.org for the few sprinkles of interactivity throughout the site.
 
-First, make sure that you are running Node 14+. If you're not sure of how to manage your node versions, see [nvm](https://github.com/nvm-sh/nvm).
+First, make sure that you are running Node 14+. If you're not sure of how to manage your node versions, see [nvm](https://github.com/nvm-sh/nvm) or [nvm-windows](https://github.com/coreybutler/nvm-windows).
 
 Then build the tutorial and doc pages (after you've installed all dependencies with `yarn install`):
 
--- a/README.md
+++ b/README.md
@@ -10,6 +10,8 @@ Other documentation is pulled from various READMEs in the main [redwoodjs/redwoo
 
 This codebase is built with https://cameronjs.com and relies on plain HTML pages and Javascript with a couple helpers built in to abstract things like partials and layouts. We use https://stimulusjs.org for the few sprinkles of interactivity throughout the site.
 
+First, make sure that you are running Node 14+. If you're not sure of how to manage your node versions, see [nvm](https://github.com/nvm-sh/nvm).
+
 Then build the tutorial and doc pages (after you've installed all dependencies with `yarn install`):
 
     yarn build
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -10,6 +10,8 @@ Other documentation is pulled from various READMEs in the main [redwoodjs/redwoo
 
 This codebase is built with https://cameronjs.com and relies on plain HTML pages and Javascript with a couple helpers built in to abstract things like partials and layouts. We use https://stimulusjs.org for the few sprinkles of interactivity throughout the site.
 
+First, make sure that you are running Node 14+. If you're not sure of how to manage your node versions, see [nvm](https://github.com/nvm-sh/nvm).
```
Tobbe, line 13: Maybe mention https://github.com/coreybutler/nvm-windows for windows users. Unfortunately I haven't used it myself, so I can't vouch for it.

docs/typescript.md
```diff
@@ -50,12 +52,24 @@ Create a `./web/tsconfig.json` file:
...
+#### Getting types for `jest` in test files
+
+If you're adding tests, you'll want to include the types for `jest` in your `tsconfig`.
+
+```diff
+-"types": []
++"types": ["jest"]
+```
+
+Currently, these are added to `node_modules` by `@redwoodjs/core` and the above approach should just work. If this is not the case, you can `npm i -D @types/jest` in the `web` folder and they will resolve.
```
Tobbe, line 73: Should we use yarn instead of npm here?

Tobbe:
@msutkowski Thanks for this! Great input 👍 I just left a couple of comments. @peterp can you have a look as well plesae?

README.md
msutkowski: I can vouch that it works as I do use it myself, but I figured someone would've made their way down that rabbit hole. 

Just an FYI, the only major difference is that the node version is set globally when using node-windows, unlike normal nvm where it's per shell 😓

docs/typescript.md
msutkowski: Noted and updated!

msutkowski:
> @msutkowski Thanks for this! Great input 👍 I just left a couple of comments. @peterp can you have a look as well please?

Just to follow up on this - I went down the 🐰 🕳️ of the TS repo trying to figure out what _actually happens_ with this change... and long story short is that it probably doesn't matter. Any TS user is going to include additional types (as shown with the Jest comment) , but maybe I'm wrong :D . If I am, please let me know, thanks! :tada:

docs/typescript.md
Burnsy: also might be worth saying `tsconfig` in web or api as we make two and you want to make sure they put it in the right place.

msutkowski: @Burnsy Thanks for the feedback! This is under the WEB section, but I guess it wouldn't hurt to add additional clarification language? 

It might be a good idea to just add in a quick cli generator for tsconfigs that adds to the root and extends in the respective directories for overrides? I'm not sure if that's already a thing in the TS support meta tracker though.

peterp (approved):

PR #525, opened 2021-01-21
Add note about starting with your own repo vs. example repo

cannikin:
@pi0neerpat what do you think about this? It still explains how to handle adding Tailwind but adds a little more context about what you'll be missing out on if you use your own instead of starting from the example.


pi0neerpat:
Looks good. I think this disclaimer will catch most problems

PR #526, opened 2021-01-25
Documentation: Forms - Select Field Documentation

mbush92:
This PR addresses Issue #[1517](https://github.com/redwoodjs/redwood/issues/1517) **Documentation: Forms - Document use of SelectField** and adds documentation to the Forms document.

### DIFF
--- a/docs/form.md
+++ b/docs/form.md
@@ -265,7 +265,7 @@ The `style` and `className` that should be passed to the HTML `<label>` tag that
 
 ## `<SelectField>`
 
-Generates an HTML `<select>` field and allows the user to select a value from the field.  Validation and error can be performed as the field is registerd with `react-hook-form`.  It is also possible to select multiple values from the field using the `multiple` attribute.  When the multiple value attribute is `true` then the return from this field will be an array of values returned in the same order as the list of options, not in the order they were selected.
+Generates an HTML `<select>` field and allows the user to select a value from the field.  Validation and error can be performed as the field is registered with `react-hook-form`.  It is also possible to select multiple values from the field using the `multiple` attribute.  When the multiple value attribute is `true` then the return from this field will be an array of values returned in the same order as the list of options, not in the order they were selected.
 
 ```html
 <SelectField name="name" validation={{required:true}}>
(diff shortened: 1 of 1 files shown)


docs/form.md
```diff
@@ -257,12 +257,116 @@ Besides the attributes listed below, any additional attributes are passed on as
...
+  <option>"pickle"</option>
+  <option>"cheese"</option>
+</SelectField>  
+
+<!-- The the user chose the lettuce, tomato and cheese options the onSubmit handler will receive { toppings:["lettuce", "tomato", "cheese"] } -->
+```
+
+#### validation
+
+Options that define how this field should be validated. The options are passed to the underlying `register` function provided by `react-hook-form`. The full list of possible values can be found in the [react-hook-form docs](https://react-hook-form.com/api#register) (ignore the usage of `ref` as that is called automaticaly for you by Redwood).
```
dthyresson, line 317: This is very nit-picky, but I believe it's preferred to refer to the framework as `RedwoodJS` especially in official docs -- even though colloquially it's "Redwood".

dthyresson (changes requested):
@mbush92 This documentation in excellent and your examples make it clear how to use and will be helpful -- which is so important in an area like forms that frankly can be confusing. 

Just one tiny change and let's merge it in some people can benefit from this.

Thanks!

docs/form.md
mbush92: I had copied that section from the `TextField` section.  I have no issue adjusting the wording, the question is should we adjust it throughout the entirety of the doc?  There are 11 instances of the usage of `Redwood` and only 4 instances of the usage as `RedwoodJS`.  Easy enough to update the full document, just want to make sure that's what you would like.

dthyresson: Am I being silly here @thedavidprice or should we try to refer to the framework more officially as RedwoodJS vs just "
"Redwood" especially in documentation?

If so @mbush92 can update all instances.

mbush92: So I did a quick search, not including this file the framework is referred to `Redwood` 140 times versus only 16 times as `RedwoodJS`. @thedavidprice and @dthyresson if we decide to make the change in nomenclature I can open another PR that addresses just that change in all of the other documents as well.

dthyresson: Yeah - I see this now. 

> Redwood is an opinionated, full-stack, serverless web application framework that will allow you to build and deploy JAMstack applications with ease.

Let's table this issue and then we can address later.

I'm going to approve and merge.

Thanks for walking down that road with me. Appreciate it.

dthyresson (approved):
LGTM and thanks again!

PR #527, opened 2021-01-25
Add a step in the RBAC section to ensure the success of refetchQuery action after deletion

### LINKED ISSUE #523: Redwood Tutorial Part 2: RBAC delete hits 404 on refetchQueries
In Part 2 of the tutorial in the RBAC sections we are asking followers to delete comments, 
```
const [deleteComment] = useMutation(DELETE, {
    refetchQueries: [
      {
        query: CommentsQuery,
        variables: { postId: comment.postId },
      },
    ],
  })
```
  
  
while this does delete the comment. The refetchQuries will fail. 
It fails because nowhere in the tutorial do we have the follower have `postId` returned in the `CommentsCell` so this variable `comment.postId` is `undefined`

The fix is to mention adding the `postId` to the `CommentsCell`.

```
export const QUERY = gql`
  query CommentsQuery($postId: Int!) {
    comments(postId: $postId) {
      id
      postId
      name
      body
      createdAt
    }
  }
````

Additionally, we will need to update the `CommentsCell.mock.js` with the `postId`

### PR DESCRIPTION
**Pull request by zainfathoni**
1. Add a step to ensure the success of `refetchQuery` action after the deletion. (fixes #523)
    Screenshot 2021-01-26 at 1 06 50 AM
    Screenshot 2021-01-26 at 1 08 03 AM
2. ~~Add `mockGraphQLQuery()` example while fixing the **BlogPost** Storybook error~~ (potentially an invalid issue)
    Screenshot 2021-01-26 at 12 44 57 AM
    Screenshot 2021-01-26 at 12 56 00 AM
">

### DIFF
--- a/TUTORIAL2.md
+++ b/TUTORIAL2.md
@@ -2335,6 +2335,26 @@ const Comment = ({ comment }) => {
 export default Comment
 ```
 
+Don't forget to update the `CommentsQuery` we're importing from **CommentsCell** to include the `postId` field, since we are relying on it to perform the `refetchQuery` after a successful deletion.
+
+```javascript{11}
+// web/src/components/CommentsCell/CommentsCell.js
+
+import Comment from 'src/components/Comment'
+
+export const QUERY = gql`
+  query CommentsQuery($postId: Int!) {
+    comments(postId: $postId) {
+      id
+      name
+      body
+      postId
+      createdAt
+    }
+  }
+`
+```
+
 Click "Delete" (as a moderator) and the comment should be removed!
 
 Ideally we'd have both versions of this component (with and without the "Delete" button) present in Storybook so we can iterate on the design. But there's no such thing as "logging in" in Storybook and our code depends on being logged in so we can check our roles...how will that work?

### CODE REVIEW

**Comment by dthyresson** [comment]
Thanks so much for this update, @zainfathoni! 

This cookbook was written up before the refetch changes Storybook was widely available, so this so so great to have for a more complete example.

I'll review more later today, but looks excellent on a first pass.

**Comment by cannikin** [comment]
@dthyresson This is in Tutorial 2, not the cookbook!

@zainfathoni Thanks for the `postId` fix! 

The second change, using `mockGraphQLQuery()`, actually shouldn't be necessary: if a Cell is rendering in Storybook, even nested inside of another component, it should use its associated mock files automatically. We've got the comments mocks created in https://redwoodjs.com/tutorial2/multiple-comments#storybook so this *should* magically work.

I've got a test repo where I've gone through the second tutorial and this is working for me... my **Full** story looks like:

```javascript
export const full = () => {
  return <BlogPost post={POST} />
}
```

And `comments.mock.js`:

```javascript
export const standard = (/* vars, { ctx, req } */) => ({
  comments: [
    {
      id: 1,
      postId: 1,
      name: 'Rob Cameron',
      body: 'First comment',
      createdAt: '2020-01-02T12:34:56Z',
    },
    {
      id: 2,
      postId: 2,
      name: 'David Price',
      body: 'Second comment',
      createdAt: '2020-02-03T23:00:00Z',
    },
  ],
})
```

And is rendering just fine:

image

You can see my sample repo in the `rc-tutorial2a` branch here: https://github.com/redwoodjs/redwood-tutorial/tree/rc-tutorial2a

If your code is in a similar state but the Cell is *not* using that mock then something bad is happening, let us know!

dthyresson:
@cannikin > This is in Tutorial 2, not the cookbook!

Yeah, I realized that 10 seconds after posting 🤦‍♂️

Sorry for the confusion and I took myself off the reviewers list.

zainfathoni:
@cannikin I see. I tried checking out [your branch](https://github.com/redwoodjs/redwood-tutorial/tree/rc-tutorial2a) and it seems working perfectly. Maybe I missed something when I was walking through the tutorial. 😅 
Anyway, let's exclude this issue for now, then. I just removed my commit for that issue, so now this PR only contains the `postId` fix.

I will drill down further on my local to see what went wrong with my implementation so far.
If it is really a problem that we need to pay attention to, I will raise another issue in this repository. 😁 

Thanks for the prompt response @cannikin! 🙏

cannikin:
Thank YOU! :)

PR #533, opened 2021-01-28
[0.24.0] Change depracated generate docs

dac09:
Docs update related to change in https://github.com/redwoodjs/redwood/pull/1685


dac09:
Sorry @thedavidprice looks like my editor is formatting the markdown files. Is there a specific plugin / setting I should be using?

PR #534, opened 2021-01-29
[0.24.0] Authentication: Azure Active Directory

jeliasson:
Updating docs for Azure Active Directory authentication to be released with `0.24.0`.

#### Related PR
https://github.com/redwoodjs/redwood/pull/1675

#### Related Issue
[redwoodjs/redwood#1672](https://github.com/redwoodjs/redwood/issues/1672#issuecomment-769766607)

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -272,9 +272,9 @@ const azureActiveDirectoryClient = new UserAgentApplication({
 ReactDOM.render(
   <FatalErrorBoundary page={FatalErrorPage}>
     <AuthProvider client={azureActiveDirectoryClient} type="azureActiveDirectory">
-      <RedwoodProvider>
+      <RedwoodApolloProvider>
         <Routes />
-      </RedwoodProvider>
+      </RedwoodApolloProvider>
     </AuthProvider>
   </FatalErrorBoundary>,
   document.getElementById('redwood-app')
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -250,7 +250,7 @@ Enter allowed redirect urls for the integrations, e.g. `http://localhost:8910`. 
 
 Under the _Authentication_ tab, tick `ID tokens`.
 
-This allows an application to request a token directly from the authorization endpoint. Checking Access tokens and ID tokens is recommended only if the application has a single-page architecture (SPA). [Learn more about the implicit grant flow](https://docs.microsoft.com/azure/active-directory/develop/v2-oauth2-implicit-grant-flow?WT.mc_id=Portal-Microsoft_AAD_RegisteredApps)
+This allows an application to request a token directly from the authorization endpoint. Checking Access tokens and ID tokens is recommended only if the application has a single-page architecture (SPA). [Learn more about implicit grant flow](https://docs.microsoft.com/azure/active-directory/develop/v2-oauth2-implicit-grant-flow?WT.mc_id=Portal-Microsoft_AAD_RegisteredApps).
 
 #### Authority
 
@@ -300,7 +300,7 @@ See [loginPopup](https://pub.dev/documentation/msal_js/latest/msal_js/UserAgentA
 
 #### getToken Options
 
-Options in method `getToken(options?)` is of type [AuthRequest](https://pub.dev/documentation/msal_js/latest/msal_js/AuthRequest-class.html). By default, method `getToken` will be called with scope `[]` which would result in `openid` and `profile` being added as described above. Furthermore, because Azure Active Directory apply [incremental consent](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/resources-and-scopes.md#dynamic-scopes-and-incremental-consent) we can extend our permissions. In example below we extend permissions by including `Mail.Read` to the scope, which extends the previous login scope example.
+Options in method `getToken(options?)` is of type [AuthRequest](https://pub.dev/documentation/msal_js/latest/msal_js/AuthRequest-class.html). By default, `getToken` will be called with scope `['openid', 'profile']`. As Azure Active Directory apply [incremental consent](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/resources-and-scopes.md#dynamic-scopes-and-incremental-consent), we can extend the permissions from the login example by including another scope, for example `Mail.Read`.
 
 ```js
 await getToken({
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -240,11 +240,7 @@ yarn add msal
 
 #### Setup
 
-To get your application credentials, create an [App Registration](https://docs.microsoft.com/en-us/azure/active-directory/develop/quickstart-register-app) in your Azure Active Directory tenant. Take a note of your generated _Application (client) ID_ and the _Directory (tenant) ID_.
-
-##### Supported account types
-
-In most cases you want to choose _Accounts in this organizational directory only (Single tenant)_, as this will allow only users in your Azure Active Directory tenant to login to your application. If you want to enable Microsoft accounts to be able to login, choose the bottom alternative.
+To get your application credentials, create an [App Registration](https://docs.microsoft.com/en-us/azure/active-directory/develop/quickstart-register-app) in your Azure Active Directory tenant. Take a note of your generated _Application ID_ (client), and the _Directory ID_ (tenant).
 
 ##### Redirect URIs
 
@@ -262,7 +258,6 @@ The Authority is a URL that indicates a directory that MSAL can request tokens f
 
 ```js
 // web/src/index.js
-import { AuthProvider } from '@redwoodjs/auth'
 import { UserAgentApplication } from 'msal'
 
 const azureActiveDirectoryClient = new UserAgentApplication({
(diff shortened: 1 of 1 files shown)


docs/authentication.md
```diff
@@ -290,9 +285,32 @@ ReactDOM.render(
 
 To setup your App Registration with custom roles and have them exposed via the `roles` claim, follow [this documentation](https://docs.microsoft.com/en-us/azure/active-directory/develop/howto-add-app-roles-in-azure-ad-apps).
 
-#### Login and Logout Options
+#### Login Options
+
+Options in method `logIn(options?)` is of type [AuthRequest](https://pub.dev/documentation/msal_js/latest/msal_js/AuthRequest-class.html) and is a good place to pass in optional [scopes](https://docs.microsoft.com/en-us/graph/permissions-reference#user-permissions) to be authorized. By default, MSAL set `scopes` to [/.default](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-permissions-and-consent#the-default-scope) which is built in for every application that refers to the static list of permissions configured on the application registration.
```
Tobbe, line 290:
```suggestion
Options in method `logIn(options?)` is of type [AuthRequest](https://pub.dev/documentation/msal_js/latest/msal_js/AuthRequest-class.html) and is a good place to pass in optional [scopes](https://docs.microsoft.com/en-us/graph/permissions-reference#user-permissions) to be authorized. By default, MSAL sets `scopes` to [/.default](https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-permissions-and-consent#the-default-scope) which is built in for every application that refers to the static list of permissions configured on the application registration.
```

Can some native English speaker please check on this. Should it be "set" or "sets"?

docs/authentication.md
```diff
@@ -290,9 +285,32 @@ ReactDOM.render(
...
+  scopes: ['User.Read.All'],
+})
+```
+
+See [loginPopup](https://pub.dev/documentation/msal_js/latest/msal_js/UserAgentApplication/loginPopup.html) or [full class documentation](https://pub.dev/documentation/msal_js/latest/msal_js/UserAgentApplication-class.html#constructors) for more information.
+
+#### getToken Options
+
+Options in method `getToken(options?)` is of type [AuthRequest](https://pub.dev/documentation/msal_js/latest/msal_js/AuthRequest-class.html).
+By default, method `getToken` will be called with options below. Essentially this is MSAL's way of using the same scope that was used in the `logIn` request.
```
Tobbe, line 303:
```suggestion
By default, method `getToken` will be called with the options below. Essentially this is MSAL's way of using the same scope that was used in the `logIn` request.
```

docs/authentication.md
```diff
@@ -262,7 +258,6 @@ The Authority is a URL that indicates a directory that MSAL can request tokens f
 
 ```js
 // web/src/index.js
-import { AuthProvider } from '@redwoodjs/auth'
```
Tobbe, line 265: Why did you remove this?

docs/authentication.md
```diff
@@ -290,9 +285,32 @@ ReactDOM.render(
...
+See [loginPopup](https://pub.dev/documentation/msal_js/latest/msal_js/UserAgentApplication/loginPopup.html) or [full class documentation](https://pub.dev/documentation/msal_js/latest/msal_js/UserAgentApplication-class.html#constructors) for more information.
+
+#### getToken Options
+
+Options in method `getToken(options?)` is of type [AuthRequest](https://pub.dev/documentation/msal_js/latest/msal_js/AuthRequest-class.html).
+By default, method `getToken` will be called with options below. Essentially this is MSAL's way of using the same scope that was used in the `logIn` request.
+
+```js
+await getToken({
+  scopes: [process.env.AZURE_ACTIVE_DIRECTORY_CLIENT_ID],
```
Tobbe, line 307: Should the client id really be passed as a scope?

jeliasson: Hmm, I think I scrolled down in the document and realized that MagicLink and Firebase do not reference these in the docs (I guess for brevity, the same as `FatalErrorBoundary` and `Redwood(Apollo)Provider`). However, when reading your comment I realize that it's there on the remaining providers.

I'll re add it for now. 😊

docs/authentication.md
jeliasson: Hmm, I think I scrolled down in the document and realized that MagicLink and Firebase do not reference these in the docs (I guess for brevity, the same as FatalErrorBoundary and Redwood(Apollo)Provider). However, when reading your comment I realize that it's there on the remaining providers.

I'll re add it for now. 😊

docs/authentication.md
jeliasson: From my understanding, passing the client id will translate into default scopes of `['openid', 'profile']` or re-use the scopes that was used in the login methods (`loginPopup` or `loginRedirect`) without being explicit about it. Unfortunately the documentation about this is very slim and the majority of their examples explicitly write out the same scope as their login scope(s).

We can consider hard-coding `['openid', 'profile']` to make this more tangible, but run the risk of missing out on future defaults without updating, or hard-code `/.defaults` but this would scope _all_ permissions on that app.

**Special OIDC Scopes behavior cases**
https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-core/docs/scopes.md#special-oidc-scopes-behavior-cases

**Related**
https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/2536

Let me know your thoughts.

docs/authentication.md
jeliasson: Thanks for spotting that!

docs/authentication.md
jeliasson: I'm pretty sure that it is _sets_ as you suggested, but let's keep this open for a bit.

docs/authentication.md
dthyresson: I read the Special OIDC behavior above and it says that the

> ClientId is removed from the scopes array when it is the only scope in the configuration. If it is not the only scope, it is treated as a resource scope and will be sent in the final server request.

So, it is important that it be the *only* scope 

>    { scopes: ['YOUR_CLIENT_ID'] } // becomes { scopes: ['openid', 'profile'] } before the request is sent (ClientId is spliced out)

otherwise:

>    { scopes: ['YOUR_CLIENT_ID', 'User.Read'] } // becomes { scopes ['YOUR_CLIENT_ID', 'User.Read', 'openid', 'profile']} before the request is sent (ClientId is treated as resource scope and therefore not spliced out)

>    { scopes: ['YOUR_CLIENT_ID', 'openid'] } // becomes { scopes ['YOUR_CLIENT_ID', 'openid', 'profile']} before the request is sent

So it should be fine, but perhaps a comment noting if any other scopes are included on that line 308 it will also be sent (if that's something one would want, I do not know).

Tobbe: > ClientId is removed from the scopes array when it is the only scope in the configuration.

To me that sounds as though we could just as well only have this (assuming `{ scopes: [] }` is the default)
```js
await getToken()
``` 

Am I right? And, if so, do we want to simplify the code, or do we want to keep it explicit?

dthyresson: @Tobbe You might be right b/c the docs say:

> For this reason, whether or not the developer adds the openid or profile scopes to their request configuration, msal@1.x will make sure they are included before sending the request to the STS.

That is, `'openid', 'profile'` will *always* be included as scopes.

Therefore, 

`await getToken()`

seems like it would be sufficient.

jeliasson: You guys are right. I believe I saw a comment that the `client_id` will use `['openid', 'profile']` _or_ the scopes that was defined in the login request. However, I can't find this reference anymore.

> For this reason, whether or not the developer adds the openid or profile scopes to their request configuration, [msal@1.x](mailto:msal@1.x) will make sure they are included before sending the request to the STS.

As this is the case, we don't need to scope to `['<client_id>']` as default but rather `[]`. 
We would still want to have `options?` to allow overriding this, right? Otherwise I'm not sure how one would extend the scope if not for below.
```js
await getToken({
  scopes: ['User.ReadWrite.All'] // becomes ['User.ReadWrite.All', 'openid', 'profile']
})
```

I'll make some tests and get back to you. In the meantime, feel free to elaborate on how the docs should be written.

jeliasson: > You guys are right. I believe I saw a comment that the `client_id` will use `['openid', 'profile']` _or_ the scopes that was defined in the login request. However, I can't find this reference anymore.

They apply incremental consent as described below and thus confirms that we don't need to pass in client id as default. 
https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-browser/docs/resources-and-scopes.md#dynamic-scopes-and-incremental-consent

jeliasson: > 
>
> > ClientId is removed from the scopes array when it is the only scope in the configuration.
> 
> To me that sounds as though we could just as well only have this (assuming `{ scopes: [] }` is the default)
> 

@Tobbe @dthyresson 
Hmm... passing/default to a empty scopes (i.e. `{ scopes: [] }`) does not work. No errors, no sign in... 🤔

docs/authentication.md
dthyresson: :+1: sets

docs/authentication.md
Tobbe: > Hmm... passing/default to a empty scopes (i.e. { scopes: [] }) does not work. No errors, no sign in... 🤔
What do we say of having openid and profile as defaults?

What?! That's like totally contradictory to their docs, no? What if you pass an empty object, or nothing at all?

jeliasson: > What?! That's like totally contradictory to their docs, no? 

I have not seen any documentation on wherever `scopes` can be null or empty array, and it does not look like it. 

> What if you pass an empty object, or nothing at all?

No, no luck. I have tested the following;

✅ **client_id**
```js
const authRequest = options || {
  scopes: [process.env.AZURE_ACTIVE_DIRECTORY_CLIENT_ID], 
}
```

✅ **default openid and profile**
```js
const authRequest = options || {
  scopes: ['openid', 'profile'], 
}
```

❌ **empty array**
```js
const authRequest = options || {
  scopes: [], 
}
```

❌ **null scopes**
```js
const authRequest = options || {
  scopes: null, 
}
```

❌ **empty object**
```js
const authRequest = options || {}
```

❌ **null**
```js
const authRequest = options || null
```

I think that defaulting to `['openid', 'profile']` is the way to go here to be honest.

Tobbe: > I think that defaulting to ['openid', 'profile'] is the way to go here to be honest.

👍 Makes it very clear exactly what scopes are used

jeliasson:
@dthyresson @Tobbe 
Please review https://github.com/redwoodjs/redwood/pull/1703 and [updated docs](https://deploy-preview-534--redwoodjs.netlify.app/docs/authentication#azure-active-directory) at your convenience. 

Also, I would like to squash all my commits in this feature branch into one. Any git-ninja that knows how to do this? 

/cc @thedavidprice

Tobbe:
> Also, I would like to squash all my commits in this feature branch into one. Any git-ninja that knows how to do this?

I think we can do that when we merge
image

Otherwise you can do it locally with something like `git rebase -i` and then force pushing, `git push -f`

thedavidprice:
All our repos have a default setting to "Squash and Merge". All good!
Screen Shot 2021-02-01 at 3 21 46 PM

Thanks, gents!

PR #535, opened 2021-01-30
Update Self-hosting_Redwood.md

JamesHighsmith:
running 
`yarn add -D pm2`
gives the errror:
`error Running this command will add the dependency to the workspace root rather than the workspace itself, which might not be what you want - if you really meant it, make it explicit by running this command again with the -W flag (or --ignore-workspace-root-check).`


jtoar:
@JamesHighsmith thanks!

PR #544, opened 2021-02-05
Fix typo in post body text

agiannelli:
I struggled with tests failing and noticed I had copied the POST value from this section. The missing L was the culprit!


Tobbe:
Thank you @agiannelli!

PR #545, opened 2021-02-05
Update TUTORIAL.md

adriatic:
The sentence:
"That's exactly what layouts do: they wrap your pages in a component that then renders the page as its child" is confusing. 

The component renders the page - which page?

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -300,7 +300,7 @@ As a world-class developer you probably saw that copy-and-pasted `<header>` and 
 
 One way to solve the `<header>` dilemma would be to create a `<Header>` component and include it in both `HomePage` and `AboutPage`. That works, but is there a better solution? Ideally there should only be one reference to the `<header>` anywhere in our code.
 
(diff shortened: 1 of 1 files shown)


adriatic:
Correct and easy to understand documentation is absolutely critical

cannikin:
It's been a while since I wrote this but I think the gist was that although the same layout is wrapping multiple pages, it only renders a single page at a time. Reading "renders the pages as its children" implies to me that it's rendering all of them together. What do you think about making the *first* instance of "pages" singular instead:

> That's exactly what layouts do: they wrap a page in a component that then renders the page as its child.

adriatic:
That is precisely what I would have done without the ability to consult the author. I would also add your extended clarification, in order to establish the connection with the graphic

image

Thank you for considering this little detail relevant to spend some time to address it

cannikin:
I've re-written this text and made a more direct reference to the diagram in an upcoming update for 0.25: #549 

I'm going to close this one, thanks for pointing it out!

adriatic:
There is really no need for merging any patches - if anyone, @cannikin knows how to fix this. Thank you Rob.

PR #551, opened 2021-02-11
Update Role-based_Access_Control.md

svenhanssen:
The examples did not reflect the code sample. This makes them in sync.

### DIFF
--- a/cookbook/Role-based_Access_Control.md
+++ b/cookbook/Role-based_Access_Control.md
@@ -458,13 +458,13 @@ Use `requireAuth()` in your services to check that a user is logged in, whether 
 It checks for a single role:
 
 ```js
-requireAuth({ role: 'editor' })
+requireAuth({ roles: 'editor' })
 ```
 
 or multiple roles:
 
 ```js
-requireAuth({ role: ['admin', 'author', 'publisher'] })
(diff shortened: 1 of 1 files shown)


thedavidprice:
Thanks @svenhanssen! Looping in @dthyresson 

And welcome to Redwood 🚀

dthyresson:
Thanks @svenhanssen.

I thin the example and the auth template got a little out of sync.

```
/**
 * Use requireAuth in your services to check that a user is logged in,
 * whether or not they are assigned a role, and optionally raise an
 * error if they're not.
 *
 * @param {string=} roles - An optional role or list of roles
 * @param {string[]=} roles - An optional list of roles

 * @example
 *
 * // checks if currentUser is authenticated
 * requireAuth()
 *
 * @example
 *
 * // checks if currentUser is authenticated and assigned one of the given roles
 * requireAuth({ role: 'admin' })
 * requireAuth({ role: ['editor', 'author'] })
 * requireAuth({ role: ['publisher'] })
 */
export const requireAuth = ({ role } = {}) => {
```

params vs the actual param and that's probably legacy (at one time it accepted just a single role).

I'll update the template to reflect.

Thanks!

dthyresson (commented):
Appproved!

PR #558, opened 2021-02-12
Remove unnecessary paragraph

mikebobadilla:
I wasn't asked to name my migration and I remember hearing that Prisma just needs one migrations file now. 

If this doesn't need to be removed, no big deal.

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -535,8 +535,6 @@ That was simple. Now we'll want to snapshot this as a migration:
 >
 > From now on we'll use the shorter `rw` alias instead of the full `redwood` argument.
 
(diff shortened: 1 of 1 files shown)


thedavidprice:
@mikebobadilla! This behavior did change in yesterday's release. So the paragraph should stay as is. 

Previously we set a default migration name in `rw db save` unless the user passed `--name`. Now, with the new Prisma CLI and command `rw prisma migrate dev`, we use the default behavior which is asking for a migration name. You can still pass `--name` to skip the interactive. Also, if you just hit <return> it'll keep you moving forward.

Always a pleasure to see you in the PRs, _especially_ paying attention to the details. Keep it up! 😄 

Screen Shot 2021-02-12 at 6 28 49 AM

PR #562, opened 2021-02-12
[v0.26.0] Add Nhost to Auth Docs

nunopato:
- Adds Nhost to Auth docs

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -10,6 +10,7 @@
 - [Firebase's GoogleAuthProvider](https://firebase.google.com/docs/reference/js/firebase.auth.GoogleAuthProvider)
 - [Ethereum](https://github.com/oneclickdapp/ethereum-auth)
 - [Supabase](https://supabase.io/docs/guides/auth)
+- [Nhost](https://docs.nhost.io/auth)
 - Custom
 - [Contribute one](https://github.com/redwoodjs/redwood/tree/main/packages/auth), it's SuperEasy™!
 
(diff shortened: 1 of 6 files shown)


docs/authentication.md
```diff
@@ -505,6 +506,41 @@ To complete setup, you'll also need to update your `api` server manually. See ht
...
+
+The following CLI command will install required packages and generate boilerplate code and files for Redwood Projects:
+
+```terminal
+yarn rw setup auth nhost
+```
+
+#### Setup
+
+Update your .env file with the following settings which can be found on your project's dashboard.
```
Tobbe, line 523:
```suggestion
Update your .env file with the following setting which can be found on your Nhost project's dashboard.
```

Tobbe (approved):

Tobbe:
This looks great @nunopato. Thanks for writing the docs.
We'll merge this when the auth provider is included in a release

Tobbe:
Actually just remembered one thing. Did you want to mention `skipFetchCurrentUser` and tell people to add that? And also why they should add it?

nunopato:
@Tobbe I am not sure, to be honest. The query for `getCurrentUser`, found [here](https://github.com/redwoodjs/redwood/blob/main/packages/auth/src/AuthProvider.tsx#L134), as expected, was failing for me because I pointed `apiProxyPath` to a Hasura endpoint on Nhost after removing the `./api` folder. It is not related to the auth provider, but at the same time, I don't understand why would the user be fetched from `apiProxyPath` when using any of the auth providers. I might be missing something here.

Tobbe:
> I don't understand why would the user be fetched from `apiProxyPath` when using any of the auth providers.

We do this so our users can augment the info they get from their auth provider with info from their own database. Like if all you get from your auth provider is an email, maybe you have a table in your own db with name and roles connected to that email. `getCurrentUser` on the API side lets you fetch that from the db, and return that to the frontend. Here's an example from the code that gets generated when you run the auth setup

```javascript
export const getCurrentUser = async ({ email }) => {
  return await db.user.findUnique({ where: { email } })
}
```

nunopato:
@Tobbe thank you for the explanation, then I think we are good to go since that is not specific to Nhost used as an auth provider.

Tobbe:
> @Tobbe thank you for the explanation, then I think we are good to go since that is not specific to Nhost used as an auth provider.

So, if you use Nhost auth together with everything else on Nhost (hasura etc), you probably want `skipFetchCurrentUser`. But if, for some reason, you're only using Nhost for auth, and have everything else setup "the Redwood way" with apollo server and prisma you probably do want to call `getCurrentUser`.

I think a note about this in the docs would be great 👍

nunopato:
@Tobbe thank you. Added the missing bit.

docs/authentication.md
```diff
@@ -505,6 +506,52 @@ To complete setup, you'll also need to update your `api` server manually. See ht
...
+
+For the docs on Authentication, see: <https://docs.nhost.io/auth>
+
+If you are also using Nhost as your GraphQL API server, you will need to set `skipFetchCurrentUser` when creating the Nhost client, as follows:
+
+```js
+const nhostClient = createClient({
+  baseURL: 'https://backend-REPLACE.nhost.app',
+  autoLogin: false,
+  skipFetchCurrentUser: true,
```
Tobbe, line 548: Sorry if I was unclear, but this is actually not how that setting is used. Instead you're supposed to add it as a prop to `<AuthProvider>`

Like this

```tsx
<AuthProvider client={nhost} type="nhost" skipFetchCurrentUser>
```

nunopato: @Tobbe I am sorry, this is what happens when you do things late at night. What a miss 🙄

Tobbe: Looking good now 👍

Tobbe (approved):

Tobbe:
As soon as https://github.com/redwoodjs/redwood/pull/1725 is part of a RW release we can merge this

thedavidprice (approved):
Thanks! I have this on the list to merge with v0.26.0 release.

PR #569, opened 2021-02-14
Tutorial: Match current `prisma migrate dev` behavior

Tobbe:
`yarn rw prisma migrate dev` never asks for a name for the migration.

This updates the tutorial to match the current behavior to not confuse any readers. When the command asks for a name again, this commit can be reverted.

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -535,9 +535,7 @@ That was simple. Now we'll want to snapshot this as a migration:
 >
 > From now on we'll use the shorter `rw` alias instead of the full `redwood` argument.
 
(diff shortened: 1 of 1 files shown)


cannikin:
Hmmmmmmmm @thedavidprice I thought you said `yarn rw prisma migrate dev` would prompt for the migration name? I double checked on Slack and was told "yes"!

Tobbe:
I think the intention was that it should ask for a name, but some of the 0.25 bugs broke that behavior.  I didn't know how long it would take to get a path release out, so I figured we could tweak the wording here in the interim.

thedavidprice:
Getting a patch release out today as the intended behavior is to ask for the migration name.

Thanks for staying on top of this! Closing out.

PR #570, opened 2021-02-14
Tutorial: Change wording around post list

Tobbe:
The original wording made it sound (to me) like I'd get an array with one or two numbers in it. This hopefully makes it more clear that it's blog post objects.


cannikin:
Ahh good call, yours is much clearer!

PR #582, opened 2021-02-19
typescript: include globals folder

Tobbe:
Needed to take advantage of this https://github.com/redwoodjs/redwood/pull/1744

### DIFF
--- a/docs/typescript.md
+++ b/docs/typescript.md
@@ -61,7 +61,7 @@ Create a `./web/tsconfig.json` file:
     "typeRoots": ["../.redwood"],
     "types": []
   },
-  "include": ["src"]
+  "include": ["src", "../.redwood/globals"]
 }
 ```
 


docs/typescript.md
```diff
@@ -61,7 +61,7 @@ Create a `./web/tsconfig.json` file:
     "typeRoots": ["../.redwood"],
     "types": []
   },
-  "include": ["src"]
+  "include": ["src", "../.redwood/globals"]
```
Tobbe, line 64: @peterp Is this the only/best way to do this? Or is there a nicer looking way to make TS find the global definitions there?

peterp: Ah! I've been looking for this message for ages.

I tried placing the types in `./node_modules/@types` which is the idiomatic and automatic way of doing this, but vscode doesn't reload those files if they are modified.

Although, I heard that if you delete the entry files, and then re-create it that vscode will pick up the changes.

peterp: Please let me know if that makes sense.

Tobbe: 1. So you want me to move the types to ./node_modules/@types and then delete+re-create the file(s)?
2. This change is about the globals, not types (which is why I'm adding them to `include` instead of `typeRoots`. I'm guessing VSCode isn't looking for code in ./node_modules/@types, so moving the globals there probably isn't an option, right?

peterp: Ah. Sorry, I got confused. I'll need to play around with this - has the globals stuff merged?

Tobbe: Yes.
> Needed to take advantage of this redwoodjs/redwood#1744

peterp: You need to add it to the `.redwood/index.d.ts` file like this:

image

You can do that using this function:
https://github.com/redwoodjs/redwood/blob/d2d0938f37a6218480554be707cde5e9451c4b96/packages/core/src/babelPlugins/babel-plugin-redwood-routes-auto-loader.ts#L98

Tobbe: I can't get that to work @peterp. 

I even tried putting the triple-slash directive directly in Routes.tsx

image

I think the triple-slash directives are only supposed to be used for types, not for global declarations. (I should probably also rename the file to just routes-globals.ts, i.e. remove .d)

peterp: Ok, odd. I believe this was working previously, and when I was writing the response to this message last night, but it was late... so my mileage may vary. Let me see if I can fix this.

Tobbe:
image

docs/typescript.md
Tobbe: Peter and I discussed on Slack and decided to keep it as it is for now

PR #583, opened 2021-02-22
Update formattedDate locales in TUTORIAL2.md

seonghyeonkimm:
### Descriptions
- When default locale is different, the docs instruction does not work as expected.
- How about setting the exact locale?


Tobbe:
The proper API to use to do this is probably https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat

cannikin:
Cool, let's try just setting the `en-US` locale and see how that works. If it ends up causing problems we can look into the `Intl` implementation.

PR #585, opened 2021-02-22
Update CLI Commands nav

### LINKED ISSUE #579: Improving the nav links in the CLI Commands doc
Some of the nav links in the [CLI Commands](https://redwoodjs.com/docs/cli-commands) doc clash (most notably #db). Maybe we should make sure all sub-command nav links are prepended with their parent command the way setup is:

https://redwoodjs.com/docs/cli-commands#setup
https://redwoodjs.com/docs/cli-commands#setup-deploy-config
https://redwoodjs.com/docs/cli-commands#setup-auth

So prisma commands would be

https://redwoodjs.com/docs/cli-commands#prisma
https://redwoodjs.com/docs/cli-commands#prisma-db

Instead of just:

https://redwoodjs.com/docs/cli-commands#db

I think I didn't do this at first just to keep things from getting too cluttered on the right, and so you could scan the command hierarchy. Thoughts @cannikin @thedavidprice?

jtoar:
> https://deploy-preview-585--redwoodjs.netlify.app/docs/cli-commands

Closes #579 by updating the right nav; subcommands are now prefixed with their parent, so there's no clashing and there's more context.

### DIFF
--- a/docs/cliCommands.md
+++ b/docs/cliCommands.md
@@ -558,9 +558,8 @@ Some generators require that their argument be a model in your `schema.prisma`. 
 | `service <name>`       | Generate a service component                                                                          |
 
 ### TypeScript generators
+
 If your project is configured for TypeScript (see [TypeScript docs](https://redwoodjs.com/docs/typescript)), the generators will automatically detect and generate `.ts`/`.tsx` files for you
-
-
 
 **Undoing a Generator with a Destroyer**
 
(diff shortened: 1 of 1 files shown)


jtoar:
@thedavidprice do you think we need to worry about redirects here? We didn't rename the entire doc, so any # links will just take them to the top instead of completely breaking.

thedavidprice:
@jtoar agreed this is an improvement. Skip the redirects. Not ideal. But not terrible enough to warrant adding the redirect overhead imho.

PR #599, opened 2021-02-26
Resolve misleading guide in Schema Relations doc

### LINKED ISSUE #595: Conflicting instructions regarding m-n relationships in schema.prisma
Hello!

I'm working with a freshly generated Redwood project, using `"@redwoodjs/core": "^0.25.1"`. I've only setup a `prisma.schema` and am generating scaffolding to go along with it. I'm using y'alls [documentation on doing relations with the Redwood generator](https://redwoodjs.com/docs/schema-relations#prisma-relations-and-redwoods-generators) and comparing that to [Prisma's docs](https://www.prisma.io/docs/concepts/components/prisma-schema/relations).

When I go about implementing [this](https://redwoodjs.com/docs/schema-relations#supported-table-structure) part of your documentation, I run into the following IDE error:

```Terminal
Error validating model "PalletProduct": Each model must have at most one id criteria. You can't have `@id` and `@@id` at the same time.
```

In my use-case I actually will benefit from defining the relating model, explicitly as Prisma refers to it. I just wanted to clear-up how I should go about resolving this?

I'm assuming either my IDE's bugging-out or I need to remove `@@id` and `@@unique` constrain my referencing columns. I can provide my schema if it'll help any, but it's pretty much a copy-pasta of the documentation.

realStandal:
Resolves #595

### DIFF
--- a/docs/schemaRelations.md
+++ b/docs/schemaRelations.md
@@ -72,7 +72,7 @@ model ProductsOnTags {
   tags      Tag[]     @relation(fields: [tagId], references: [id])
   productId Int
   products  Product[] @relation(fields: [productId], references: [id])
-  @@id([tagId, productId])
+  @@unique([tagId, productId])
 }
 ```
 
--- a/docs/schemaRelations.md
+++ b/docs/schemaRelations.md
@@ -50,6 +50,8 @@ Since there's no `id` column here, you can't use the SDL generator with the `--c
 
 ## Supported Table Structure
 
+To support both CRUD actions and to remain consistent with Prisma's m-n-relationships, a combination of [`@id`](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#id) and [`@@unique`](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#unique-1) can be used. With this, `@id` is used to create a primary key on the lookup-table and `@@unique` is used to maintain its unique index, which was previously accomplished by the primary key created with `@@id`.
+
 You can get this working by creating an explicit relationship—defining the table structure yourself:
 
 ```javascript
--- a/docs/schemaRelations.md
+++ b/docs/schemaRelations.md
@@ -42,15 +42,17 @@ These relationships can be [implicit](https://www.prisma.io/docs/concepts/compon
 
 ## CRUD Requires an `@id`
 
-CRUD (Create, Retrieve, Update, Delete) actions in Redwood currently require a single, unique field in order to retrieve, update or delete a record. Generally this is an `id` column. It doesn't have to be named `id`, but needs to be a column denoted with Prisma's `@id` syntax which marks it as a primary key in the database. This field is guaranteed to be unique and so can be used to find one specific record.
+CRUD (Create, Retrieve, Update, Delete) actions in Redwood currently require a single, unique field in order to retrieve, update or delete a record. This field must be denoted with Prisma's [`@id`](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#id) attribute, marking it as the tables's primary key. This field is guaranteed to be unique and so can be used to find a specific record.
 
-Prisma's implicit many-to-many relationship syntax creates a table _without_ a column marked as an `@id`. It uses confusingly similar token `@@id` which just creates a unique _index_ on the two columns that are foreign keys to the two tables that are being joined. The diagram above shows the result of letting Prisma create an implicit relationship.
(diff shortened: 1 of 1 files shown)


docs/schemaRelations.md
```diff
@@ -72,7 +72,7 @@ model ProductsOnTags {
   tags      Tag[]     @relation(fields: [tagId], references: [id])
   productId Int
   products  Product[] @relation(fields: [productId], references: [id])
-  @@id([tagId, productId])
+  @@unique([tagId, productId])
```
realStandal, line 75: I'm not sure if y'all want to default to the `@@unique` or leave it off and let the user decide if they want repetitive m-n relationships. With it, a `tagId` can only ever correlate to a `productId` once.

cannikin: We could add a note that `@@id` creates unique relationships by enforcing a private key, but `@@unique` effectively does the same thing, but still allowing you to define your own `@id`.

realStandal: @cannikin Sounds good, just pushed that change. I'm not sure how nit-picky you'd like to be with the docs, but `@@id` is described (above what I've been changing) as *just* creating a unique index across multiple columns. Unless I'm misunderstanding unique indexes and primary keys, it should be described as creating a primary key spanning multiple columns and the unique index is a consequence of that.

Again, it's nit-picking but I bring it up more as the question: should it be explicit what is being lost/gained or in this case rearranged by adhering to Redwood's workflow?

docs/schemaRelations.md
```diff
@@ -50,6 +50,8 @@ Since there's no `id` column here, you can't use the SDL generator with the `--c
 
 ## Supported Table Structure
 
+To support both CRUD actions and to remain consistent with Prisma's m-n-relationships, a combination of [`@id`](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#id) and [`@@unique`](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#unique-1) can be used. With this, `@id` is used to create a primary key on the lookup-table and `@@unique` is used to maintain its unique index, which was previously accomplished by the primary key created with `@@id`.
```
realStandal, line 53: Should "its unique index" be replaced with "the tables unique index"?

docs/schemaRelations.md
cannikin: Yup I agree, we should be nit-picky in this case and explain what `@@id` is really doing and that `@@unique` isn't the same thing, but is *probably* okay in most cases (and required in the case where you want to scaffold it, because you need an `@id`).

docs/schemaRelations.md
```diff
@@ -42,15 +42,17 @@ These relationships can be [implicit](https://www.prisma.io/docs/concepts/compon
...
 
 ## Supported Table Structure
 
-To support both CRUD actions and to remain consistent with Prisma's m-n-relationships, a combination of [`@id`](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#id) and [`@@unique`](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#unique-1) can be used. With this, `@id` is used to create a primary key on the lookup-table and `@@unique` is used to maintain its unique index, which was previously accomplished by the primary key created with `@@id`.
+To support both CRUD actions and to remain consistent with Prisma's many-to-many relationships, a combination of the `@id` and [`@@unique`](https://www.prisma.io/docs/reference/api-reference/prisma-schema-reference#unique-1) attributes can be used. With this, `@id` is used to create a primary key on the lookup-table; and `@@unique` is used to maintain the table's unique index, which was previously accomplished by the primary key created with `@@id`.
+
+> Removing `@@unique` would let a specific **Product** refrence a particular **Tag** more than a single time.
```
realStandal, line 55: Could be removed, not sure how confusing vs helpful it is. I imagine *most* use-cases would need the `@@unique` attribute

docs/schemaRelations.md
realStandal: @cannikin Pushed that change and changed the vocabulary to better match what Prisma uses on their docs. Thanks again for letting me take this one over, helped to reinforce Prisma for me :)

thedavidprice:
ping @cannikin

docs/schemaRelations.md
cannikin: Yeah let's leave it in so people understand the consequences!

docs/schemaRelations.md
cannikin: "its" works for me, you just mentioned "lookup-table" a few words before, I think it's okay!

cannikin:
Thanks so much for this one @realStandal!

PR #602, opened 2021-03-01
fix: add node_modules/@types to typeRoots in typescript docs

jangxyz:
This fixes [#1820](https://github.com/redwoodjs/redwood/issues/1820)

Should apply for both web and api sides.


thedavidprice:
@Tobbe would you be able to review this one? Thanks in advance!

Tobbe (approved):

PR #609, opened 2021-03-05
yet another round of pico typo fixes

AntonioMeireles:
Hey All, 

because it's Friday, 😄 

Keep the great work && a great weekend!

### DIFF
--- a/cookbook/File_Upload.md
+++ b/cookbook/File_Upload.md
@@ -492,7 +492,7 @@ Now let's just add the ability to bring back the uploader if you decide you want
 
 We're borrowing the styles from the submit button and made sure the image has both a top and bottom margin so it doesn't crash into the new button.
 
-## The Wrapup
+## The Wrap-up
 
 Files uploaded! There's plenty of ways to integrate a file picker and this is just one, but we think it's simple, yet flexible. We use the same technique on the [example-blog](https://github.com/redwoodjs/example-blog).
 
--- a/docs/webpackConfiguration.md
+++ b/docs/webpackConfiguration.md
@@ -59,7 +59,7 @@ yarn rw build --stats
 
 Note that this'll skip building the api side.
 
-This config uses [Webpack Bundle Analyzer](https://github.com/webpack-contrib/webpack-bundle-analyzer). When it finishes, it'll launch an interactive zoomable, treemap in your browser to examine the contents of all your bundles.
+This config uses [Webpack Bundle Analyzer](https://github.com/webpack-contrib/webpack-bundle-analyzer). When it finishes, it'll launch an interactive zoomable, _treemap_ in your browser to examine the contents of all your bundles.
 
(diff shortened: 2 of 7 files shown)


cookbook/File_Upload.md
```diff
@@ -216,7 +216,7 @@ const ImageForm = (props) => {
   />
 ```
 
-Well lookie here:
+Well, look here:
```
jtoar, line 219: Ok if we preserve Rob's humor here or did you find it confusing?

docs/webpackConfiguration.md
```diff
@@ -59,7 +59,7 @@ yarn rw build --stats
 
 Note that this'll skip building the api side.
 
-This config uses [Webpack Bundle Analyzer](https://github.com/webpack-contrib/webpack-bundle-analyzer). When it finishes, it'll launch an interactive zoomable, treemap in your browser to examine the contents of all your bundles.
+This config uses [Webpack Bundle Analyzer](https://github.com/webpack-contrib/webpack-bundle-analyzer). When it finishes, it'll launch an interactive zoomable, tree-map in your browser to examine the contents of all your bundles.
```
jtoar, line 62: Webpack uses "treemap" all over their docs so I feel like this one is ok.

jtoar (commented):
Thanks @AntonioMeireles, always amazed at how many typos slip through so really grateful for these! Agree with most of the fixes, just had comments about two of them, let me know what you think.

cookbook/File_Upload.md
AntonioMeireles: well, it's an idiomatic thing, and i guess that as time goes by most of the audience will end being non native english speakers.. so some linguistic _metadata_ IMHO ends lost over the wire. i guess that a middle ground would be to start using as a convention italics when we actually intend to spell something in a non canonical way. will do that in next force push if you don't mind.

docs/webpackConfiguration.md
AntonioMeireles: @jtoar 

gotcha. will also italicise it if you don't mind.

AntonioMeireles:
@jtoar 

just pushed again. see if palatable now... if not i'll just revert those two mods. 

All the best && thanks for your patience

António

cookbook/File_Upload.md
cannikin: My humor transcends all language barriers! :)

My only issue with italics is that I use them just for emphasis of certain words to mimic speech, and in this case I wouldn't emphasize "lookie" in that context.

AntonioMeireles: @cannikin (hi!), @jtoar 

so, how do we proceed here ?

just drop this mod, whatever the form ?

AntonioMeireles: well, just dropped it in latest force-push. it's friday after all :-)

PR #616, opened 2021-03-09
Use new migrate command in "migration" section of Postgres doc

thedavidprice:
`yarn rw db up` --> `yarn redwood prisma migrate dev` with instructions to remove existing migrations if applicable.

### DIFF
--- a/docs/localPostgresSetup.md
+++ b/docs/localPostgresSetup.md
@@ -136,7 +136,8 @@ yarn redwood prisma migrate dev
 If you've already created migrations using SQLite, e.g. you have a migrations directory at `api/db/migrations`, follow this two-step process.
 
 #### 1. Remove existing migrations
-**For Linux and Mac OS**
+**For Linux and Mac OS**  
+From your project root directory, run either command corresponding to your OS.
 ```bash
 rm -rf api/db/migrations
 ```
--- a/docs/localPostgresSetup.md
+++ b/docs/localPostgresSetup.md
@@ -143,7 +143,7 @@ rm -rf api/db/migrations
 
 **For Windows OS**
 ```bash
-rmdir /S api\db\migrations
+rmdir /s api\db\migrations
 ```
 
 > Note: depending on your project configuration, your migrations may instead be located in `api/prisma/migrations`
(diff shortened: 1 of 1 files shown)


docs/localPostgresSetup.md
```diff
@@ -125,19 +125,34 @@ The following components make up the base URL of your database, they are always
...
+**For Linux and Mac OS**
 ```bash
-yarn rw db save
+rm -rf api/db/migrations
 ```
 
-Both commands will create and migrate the Postgres database you specified in your `.env`.
+**For Windows OS**
+```bash
+rmdir /S api\db\migrations
```
thedavidprice, line 146:
```suggestion
rmdir /s api\db\migrations
```

docs/localPostgresSetup.md
```diff
@@ -125,22 +125,36 @@ The following components make up the base URL of your database, they are always
...
-yarn rw db up
+yarn redwood prisma migrate dev
 ```
 
-If you haven't created migrations yet, use `save`:
+### Migrate from SQLite to Postgres
+If you've already created migrations using SQLite, e.g. you have a migrations directory at `api/db/migrations`, follow this two-step process.
 
+#### 1. Remove existing migrations
+**For Linux and Mac OS**
```
thedavidprice, line 139:
```suggestion
**For Linux and Mac OS**  
From your project root directory, run either command corresponding to your OS.
```

docs/localPostgresSetup.md
```diff
@@ -125,22 +125,37 @@ The following components make up the base URL of your database, they are always
...
+From your project root directory, run either command corresponding to your OS.
 ```bash
-yarn rw db save
+rm -rf api/db/migrations
 ```
 
-Both commands will create and migrate the Postgres database you specified in your `.env`.
+**For Windows OS**
+```bash
+rmdir /s api\db\migrations
```
Tobbe, line 147: This is what I get on Windows

image

It all comes down to what shell you use.

StackOverflow says this is how you're supposed to do it if you're using PowerShell
```
Remove-Item -LiteralPath "foldertodelete" -Force -Recurse
``` 
But apparently it doesn't work for some people, so they have to do
```
Get-ChildItem -Path "foldertodelete" -Recurse | Remove-Item -force -recurse
Remove-Item "foldertodelete" -Force 
```

Further down someone says you're now able to do this in PowerShell, because apparently it now understands some unix-like syntax
```
rm -R .\DirName
```

It's a mess :( 

Here's the thread for reference https://stackoverflow.com/questions/7909167/how-to-quietly-remove-a-directory-with-content-in-powershell

thedavidprice: I give up...

¯\_(ツ)_/¯

PR #613, opened 2021-03-07
Move tutorials to learn.redwoodjs.com

clairefro:
Updates build, links and redirects to migrate tutorials pt. I and II to the "learn" subdomain. 
"Learn" is already live [here](https://learn.redwoodjs.com/), so this PR finishes the job. 

I've consolidated "Tutorial" and "TutorialII" in the nav to "Tutorials"

image

**Note**
I could not bring myself to delete the TUTORIAL.md and TUTORIAL2.md, as they are revered like the Magna Carta. 
So I've put them in an `archived` dir

PR #622, opened 2021-03-11
bump requirement to node.js >=14

thedavidprice:
We bumped the requirement in the v0.27.0 release.

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -48,7 +48,7 @@ yarn redwood upgrade
 
 During installation, RedwoodJS checks if your system meets version requirements for Node and Yarn:
 
-- node: ">=12"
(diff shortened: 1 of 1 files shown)


thedavidprice:
FYI @clairefro @cannikin

clairefro:
I've updated the English and French on learn 

We should decide soon whether we pull the trigger or the plug on learn, so we don't have to worry about syncing doc bases.
If we go for learn, is there anything else that needs to be done before implementing the .com tutorial redirects (#613)?

cannikin:
Based on our convo from the other day I don't think there's anything else? I can start tweaking the header on learn. to try and get closer to the one on .com, but that's not stopping us from linking over now!

PR #632, opened 2021-03-16
Add info + instructions to contribute to Tutorials via learn repo

thedavidprice:
@cannikin @clairefro In the past when we've moved docs around, Dom and I would leave a "redirect message" to point people to the location of the new doc. I've re-added both Tutorial.md and Tutorial2.md with info + direction to the respective directories in the lean.rwjs.com repo.

I started with one file, Tutorials.md, but then realized people might be following permalinks to the previously existing files themselves. Thus, the 2 files with identical content.

Thoughts/suggestions about what's most helpful now regarding the two files and content? I imagine we can remove these altogether in the future — just a temporary "helper".

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -1,0 +1,11 @@
+## Tutorial Contributions (and Translators) are Welcome!
+To support translations, the Redwood Tutorials have been moved to the `learn.redwoodjs.com` Repo:
+- **Tutorial Part 1:** [learn.redwoodjs.com/docs/tutorial/](https://github.com/redwoodjs/learn.redwoodjs.com/tree/main/docs/tutorial)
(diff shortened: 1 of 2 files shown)


clairefro:
Thanks David - I hadn't thought about permalinks to the Github content itself 

I think this is a great idea for those that are used to editing the tutorial but didn't notice the updated README

cannikin:
Do you guys have any plans for long to keep those files around? Seems like the entire repo would eventually just become cluttered with these stubs if we have to support them for all time...

thedavidprice:
> Do you guys have any plans for how long to keep those files around? Seems like the entire repo would eventually just become cluttered with these stubs if we have to support them for all time...

^^ we've had them before and they've naturally been removed over time. Given conversations about changes we're considering for v1, my guess is they'll be stripped out during that process.

Merging for now 'cause I think it's helpful near term. No strong opinions from me about the next steps (and when they happen).

Thanks!

PR #633, opened 2021-03-16
Provide more search results

jtoar:
Right now, for any given search, we only provides 3 results:

image

While it's better than nothing, it just isn't enough. The first few results should always be the most relevant, but we should provide a lot more. This PR changes the `hitsPerPage` from `3` to `50`. The new number, `50`, is arbitrary. Open to suggestions on exactly how many more a lot more is.

To accommodate this change, the dialog's height is capped and y-overflow is set to scroll: https://s.tape.sh/DF7FXh2b.

This isn't a replacement for actually having good search--tending to the index, etc. But, it's a fix we can make right now for the immediate better.

### DIFF
--- a/code/html/_search.html
+++ b/code/html/_search.html
@@ -9,7 +9,8 @@
     data-target="application.search"
   />
   <div
-    class="hidden absolute top-0 left-0 right-0 bg-white z-20 p-4 mt-12 rounded border border-red-200 overflow-x-auto"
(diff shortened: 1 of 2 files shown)


cannikin:
Oh man after testing it out, that 50 seems excessive...it makes the results "feel" worse to me because it's like "we don't know, here's everything, good luck!" Let's try like 10?

jtoar:
@cannikin reduced!

cannikin:
LET'S DO IT

PR #636, opened 2021-03-18
[v0.28.0] add prisma db pull command

thedavidprice:
depends on https://github.com/redwoodjs/redwood/pull/2021

Adds new prisma db pull command, which was available in Prisma v2.18 and will be supported in Redwood v0.28.

### DIFF
--- a/docs/cliCommands.md
+++ b/docs/cliCommands.md
@@ -106,7 +106,7 @@ web/src/Routes.js:17:19: error (INVALID_ROUTE_PATH_SYNTAX): Error: Route path co
 
 Launch an interactive Redwood shell (experimental):
 - This has not yet been tested on Windows.
-- The Prisma Client must be generated _prior_ to running this command, e.g. `yarn rw db up`. This is a known issue.
+- The Prisma Client must be generated _prior_ to running this command, e.g. `yarn rw prisma generate`. This is a known issue.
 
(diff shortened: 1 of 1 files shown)


docs/cliCommands.md
```diff
@@ -106,7 +106,7 @@ web/src/Routes.js:17:19: error (INVALID_ROUTE_PATH_SYNTAX): Error: Route path co
 
 Launch an interactive Redwood shell (experimental):
 - This has not yet been tested on Windows.
-- The Prisma Client must be generated _prior_ to running this command, e.g. `yarn rw db up`. This is a known issue.
+- The Prisma Client must be generated _prior_ to running this command, e.g. `yarn rw prisma generate`. This is a known issue.
```
cannikin, line 109: Did we want to use the full `redwood` keyword here instead of `rw`? I see that it was explicitly changed to the full one below...

thedavidprice: Ah, good catch. I've been making that change as I go but missed this one.

What do you think about me doing a find+replace for this entire doc — probably a good idea, yes?

cannikin: Yeah that'll be way easier!

thedavidprice: done! found a few false positives that I fixed.

PR #641, opened 2021-03-23
add example of beforeQuery export for polling

ajoslin103:
I wanted to add an example of exporting beforeQuery to turn on Apollo query polling

### DIFF
--- a/docs/cells.md
+++ b/docs/cells.md
@@ -117,18 +117,25 @@ This means you can think backwards about your Cell's props from your SDL: whatev
 
 ### beforeQuery
 
-`beforeQuery` is a lifecycle hook. The best way to think about it is as an API for configuring Apollo Client's `Query` component (so you might want to check out the [docs](https://www.apollographql.com/docs/react/api/react-components/#query) for it).
+`beforeQuery` is a lifecycle hook. The best way to think about it is as an API for configuring Apollo Client's `Query` component (so you might want to check out Apollo's [docs](https://www.apollographql.com/docs/react/api/react-components/#query) for it).
 
 By default, `beforeQuery` gives any props passed from the parent component to `Query` so that they're available as variables for `QUERY`. It'll also set the fetch policy to `'cache-and-network'` since we felt that this is the behavior users want most of the time.
 
 <!-- Source: https://github.com/redwoodjs/redwood/issues/717 -->
 ```javascript
 export const beforeQuery = (props) => {
-  return { variables: props, fetchPolicy: 'network-only' }
+  return { variables: props, fetchPolicy: 'cache-and-network' }
 }
 ```
 
-But you can of course override this by exporting your own.
+For example, if you wanted to turn on Apollo's polling option, and prevent caching, you could export something like this:
+
+<!-- Source: https://github.com/redwoodjs/redwood/issues/717 -->
+```javascript
+export const beforeQuery = (props) => {
+  return { variables, fetchPolicy: 'no-cache', pollInterval: 1000 * 2.5 }
+}
+```
 
 ### afterQuery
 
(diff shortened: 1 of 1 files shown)


docs/cells.md
```diff
@@ -117,18 +117,25 @@ This means you can think backwards about your Cell's props from your SDL: whatev
...
 }
 ```
 
-But you can of course override this by exporting your own.
+For example, if you wanted to turn on Apollo's polling option, and prevent caching, you could export something like this:
+
+<!-- Source: https://github.com/redwoodjs/redwood/issues/717 -->
+```javascript
+export const beforeQuery = (props) => {
+  return { variables: props, fetchPolicy: 'no-cache', pollInterval: 1000 * 2.5 }
```
dthyresson, line 136: From prior PR's, @peterp has had me change expressions like: `1000 * 2.5` to `2_500`.

docs/cells.md
```diff
@@ -117,18 +117,25 @@ This means you can think backwards about your Cell's props from your SDL: whatev
...
 <!-- Source: https://github.com/redwoodjs/redwood/issues/717 -->
 ```javascript
 export const beforeQuery = (props) => {
-  return { variables: props, fetchPolicy: 'network-only' }
+  return { variables: props, fetchPolicy: 'cache-and-network' }
 }
 ```
 
-But you can of course override this by exporting your own.
+For example, if you wanted to turn on Apollo's polling option, and prevent caching, you could export something like this:
```
dthyresson, line 131: How about some links to Apollo docs as well:

`[polling option](https://www.apollographql.com/docs/react/data/queries/#polling)`

and

`[prevent caching](https://www.apollographql.com/docs/react/data/queries/#setting-a-fetch-policy)`

docs/cells.md
ajoslin103: gotcha

docs/cells.md
ajoslin103: good idea!

thedavidprice:
@ajoslin103 thanks for this!

docs/cells.md
```diff
@@ -117,18 +117,25 @@ This means you can think backwards about your Cell's props from your SDL: whatev
...
-`beforeQuery` is a lifecycle hook. The best way to think about it is as an API for configuring Apollo Client's `Query` component (so you might want to check out the [docs](https://www.apollographql.com/docs/react/api/react-components/#query) for it).
+`beforeQuery` is a lifecycle hook. The best way to think about it is as an API for configuring Apollo Client's `Query` component (so you might want to check out Apollo's [docs](https://www.apollographql.com/docs/react/api/react-components/#query) for it).
 
 By default, `beforeQuery` gives any props passed from the parent component to `Query` so that they're available as variables for `QUERY`. It'll also set the fetch policy to `'cache-and-network'` since we felt that this is the behavior users want most of the time.
 
 <!-- Source: https://github.com/redwoodjs/redwood/issues/717 -->
 ```javascript
 export const beforeQuery = (props) => {
-  return { variables: props, fetchPolicy: 'network-only' }
+  return { variables: props, fetchPolicy: 'cache-and-network' }
```
cannikin, line 127: This actually sets an additional key, `nextFetchPolicy` now! We haven't edited this doc in a while... mind adding that in at the same time? 🙏

```javascript
beforeQuery = (props) => ({
  variables: props,
  fetchPolicy: 'cache-and-network',
  nextFetchPolicy: 'cache-first',
})
```

cannikin (changes requested):

docs/cells.md
ajoslin103: done! I love this framework !!

dthyresson: > I just pitched it at work this week and told 3 other groups to start looking at it :)

Great to hear!

ajoslin103 (commented):
I don't understand - it looks like I'm supposed to review?

cannikin (approved):

cannikin:
@ajoslin103 Nope, you're good, thanks!

PR #647, opened 2021-03-27
fix storybook path error

jangxyz:
`storybook.main.js` seems to be a path error. Current implemented code points to `web/config/storybook.config.js` ([packages/internal/src/paths.ts#L80](https://github.com/redwoodjs/redwood/blob/439f23142805c87a369ba1597bd4099d6ba1d473/packages/internal/src/paths.ts#L80))


thedavidprice:
Thank you!

Issue #650, opened 2021-03-28
Modify import @apollo/client redwood js.


vchoy:
Cause:
https://github.com/redwoodjs/redwood/issues/2132

Documentation needs to be changed to import directly from apollo.

thedavidprice:
out of date yarn.lock was causing this behavior. deleting it and then running `yarn install` resolved.

PR #651, opened 2021-03-28
#650: change to import directly from apollo

vchoy:


docs/graphql.md
```diff
@@ -50,7 +50,7 @@ const MutateButton = () => {
 }
 ```
 
-Note that you're free to use any of Apollo's other hooks, you'll just have to import them from `@apollo/client` instead. In particular, these two hooks might come in handy:
+Note that you're free to use any of Apollo's other hooks, you'll just have to import them from `@redwoodjs/web/node_modules/@apollo/client` instead. In particular, these two hooks might come in handy:
```
thedavidprice, line 53: The original seems correct. The workaround you found above is actually a bit circular and not something we'd recommend. In this case, it gives me an idea about what you might have going on.

thedavidprice (changes requested):

thedavidprice:
Hi @vchoy! Thanks for working through this and opening a PR! People "new to rw" are absolutely welcome.

So, you should be able to use any of [these Apollo hooks](https://www.apollographql.com/docs/react/api/react/hooks/#usequery) with the given syntax. E.g. `import { useLazyQuery } from "@apollo/client"`.

If that's not the case for you (as you've indicated), I wonder if you're working on a Redwood project that was originally created/installed prior to v0.26. We de-coupled Redwood's GraphQL Client from Apollo so that people could swap providers if they wanted. Check out the [Release Notes for v0.26](https://github.com/redwoodjs/redwood/releases/tag/v0.26.2) and especially the "How to upgrade..." section.

Does your project's web/src/App.js (formerly index.js) effectively [look like this](https://github.com/redwoodjs/redwood/blob/main/packages/create-redwood-app/template/web/src/App.tsx). Note the `import { RedwoodApolloProvider } from '@redwoodjs/web/apollo'` at the top. My hypothesis is that you're missing this from your app. (Which would break other things, but that's the best idea I have as of now.)

vchoy:
Closing PR, as this the solution as the lock file is out of sync.

Removing the yarn.lock file and getting yarn install to generate again, will resolve this issue.

PR #652, opened 2021-03-29
Typo corrections in Accessibility and Logger documentation

AntonioMeireles:
Hey all, 

subject says it all :-) 

António


dthyresson:
LGTM. Thanks for these corrections and making the docs better!

PR #660, opened 2021-04-04
Updated forms doc to match redwood pr#2167.

cjreimer:
This change adds some clarification into the docs regarding transformValue, with respect to the changes in pr redwoodjs/redwood#2167

@Tobbe flagged for you as you have helped me in the pr review.  Thanks!

### DIFF
--- a/docs/form.md
+++ b/docs/form.md
@@ -438,7 +438,7 @@ If the transformValue is set to `DateTime`, `Float`, `Int`, or `Json` and the tr
       </Form>
 ```
 
-If the number field is not modified and remains empty, it will return an `undefined` into the 'submit' function as per the below, as an empty string cannot be converted to an `integer`.
+If the <NumberField> is not modified and remains empty, it will return an `undefined` into the 'submit' function as per the below, as an empty string cannot be converted to an `integer`.
 ```
 { intField: undefined }
 ```
@@ -481,7 +481,6 @@ Caveat:  JSON validation will not applied if a custom validation function, such 
 #### validation
 
 See InputFields [validation](#inputfields-attributes)
-
 
 #### errorStyle / errorClassName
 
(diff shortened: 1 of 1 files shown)


docs/form.md
```diff
@@ -430,6 +430,26 @@ You can also pass a function to `transformValue`. For instance, you might remove
...
+```javascript
+      <Form onSubmit={submit}>
+        <NumberField name="intField" defaultValue="" transformValue="Int" />
+      </Form>
+```
+
+If the <NumberField> is not modified and remains empty, it will return an `undefined` into the 'submit' function as per the below, as an empty string cannot be converted to an `integer`.
+```
+{ intField: undefined }
+```
```
Tobbe, line 444:
```suggestion
```

docs/form.md
```diff
@@ -430,6 +430,26 @@ You can also pass a function to `transformValue`. For instance, you might remove
...
+If the production environment is set to `development` or `test`, it will also issue a console warning upon a failed transformation.  It is recommended to set up your field validation to avoid cases of failed transformation.  In the specific case of the example above, it would be recommended to add a `validation={{ required: true }}` to the code as per the below.
```
Tobbe, line 445:
```suggestion
If the `NODE_ENV` environment variable is set to `development` or `test`, it will also issue a console warning upon a failed transformation.  It is recommended to set up your field validation to avoid cases of failed transformation.  In the specific case of the example above, it would be recommended to add a `validation={{ required: true }}` to the code as per the below.
```

docs/form.md
```diff
@@ -440,6 +460,24 @@ Besides the attributes listed below, any additional attributes are passed on as
...
+      <Form onSubmit={onSubmit}>
+        <TextAreaField
+          name="jsonField"
+          transformValue="Json"
+        />
+        <Submit>Save</Submit>
+      </Form>
+```
+
+Caveat:  JSON validation will not applied if a custom validation function, such as `fcn` is provided via the prop as follows: `validation={{ validate: fcn }}`
```
Tobbe, line 479:
```suggestion
Caveat:  JSON validation will not be applied if a custom validation function, such as `fcn` is provided via the prop as follows: `validation={{ validate: fcn }}`
```

Tobbe (changes requested):

Tobbe (approved):

Tobbe:
Thanks for updating the docs Curtis. We'll merge this when we release the next version of RW

thedavidprice:
✅ on the list!

PR #667, opened 2021-04-10
Spelling Corrections in CLI Command documentation

AntonioMeireles:
Hey all! 

Have a great weekend too!

P.S. the 2nd one is a clear one. 1st one may or not be intentional. FWIW if it was i'm not too fond about it because once we start going warp speed and have things translated in multiple idioms this kind of gems get lost in translation and only add entropy to the whole process. Anyway - if it was intentional - and policy is to keep it i'll drop it :-)

### DIFF
--- a/docs/cliCommands.md
+++ b/docs/cliCommands.md
@@ -238,7 +238,7 @@ This means that `yarn redwood db generate` needs to be run after every change to
 
 ### introspect
 
-Introspect your database and generate models in `./api/db/schema.prisma`, overedwoodriting existing models.
+Introspect your database and generate models in `./api/db/schema.prisma`, overwriting existing models.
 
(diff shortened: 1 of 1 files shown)


docs/cliCommands.md
```diff
@@ -238,7 +238,7 @@ This means that `yarn redwood db generate` needs to be run after every change to
 
 ### introspect
 
-Introspect your database and generate models in `./api/db/schema.prisma`, overedwoodriting existing models.
```
dthyresson, line 241: :)

I bet that was a replace of `rw` with `redwood` since at some point it was desired that in CLI command docs the full `redwood` was preferred to the `rw` shorthand.

dthyresson:
LGTM

PR #682, opened 2021-04-30
Adding JSON headers

ryancwalsh:
It took me a while to figure out how to ensure that the response headers used 'Content-Type': 'application/json'. Maybe this addition to the docs will help others.

### DIFF
--- a/docs/serverlessFunctions.md
+++ b/docs/serverlessFunctions.md
@@ -25,6 +25,9 @@ It'll give you a stub that exports a handler that returns a status code&mdash;th
 export const handler = async (event, context) => {
   return {
     statusCode: 200,
+    headers: {
+      'Content-Type': 'application/json',
+    },
     body: JSON.stringify({
       data: '${name} function',
     }),


dthyresson:
@ryancwalsh Good point. Do you think that the Content Type should be added to the function generator template as well:

https://github.com/redwoodjs/redwood/blob/19d2bc1eee079f7de3dea8ff93de0b61d9599d80/packages/cli/src/commands/generate/function/templates/function.ts.template#L24

?

ryancwalsh:
@dthyresson Sure! I just submitted https://github.com/redwoodjs/redwood/pull/2457 for you.

dthyresson:
@ryancwalsh As part of the security, webhooks and other docs this has been added:

https://redwoodjs.com/docs/serverless-functions

If ok, let's close this and we'll look at redwoodjs/redwood#2457 f

ryancwalsh:
@dthyresson Sounds good, thanks.

PR #686, opened 2021-05-11
Remove idea of sharing package outside of Redwood

andrew-hwahin:
related to issue https://github.com/redwoodjs/redwood/issues/1888


jtoar:
@andrew-hwahin Thanks!

PR #691, opened 2021-05-13
Add a doc on Services

jtoar:
Redwood's business logic abstraction needs it's own doc, especially Secure Services coming out in the next release!

### DIFF
--- a/docs/services.md
+++ b/docs/services.md
@@ -1,41 +1,32 @@
 # Services
 
-How do you confidently write your business logic in a secure way while scaling?
+Redwood puts all your business logic in one place—Services. These can be used by your GraphQL API or any other place in your backend code. Redwood does all of the annoying stuff for you, just write your business logic!
 
 ## Secure Services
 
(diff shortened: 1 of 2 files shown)


docs/services.md
```diff
@@ -0,0 +1,369 @@
...
+REDWOOD_SECURE_SERVICES=1
+```
+
+Once you do this, you'll see your Services suddenly become inaccessible, with an error message in the console when you start your dev server:
+
+```
+Must define a `beforeResolver()` in posts/posts.js
+```
+
+Which means it worked!
```
thedavidprice, line 33:
```suggestion
Which means it worked!

> **Heads Up**: Using Env Vars for Config is a _Two-step Process_  
> Technically, you need to enable this Env Var both for your local dev environment and your production deploy environment. In this case, we are assuming you are committing your `env.defaults` file to your project Repo, which will add the Env Var to both contexts. If not, you also need to specifically add the Env Var to your hosting provider config.
```

PR #693, opened 2021-05-14
Corrects typos in security and serverless docs

AntonioMeireles:
Good Morning && All the best


dthyresson (approved):
Thanks again for these corrections!

PR #694, opened 2021-05-15
docs/testing.md: fix typo in sample code

AntonioMeireles:
subject says it all. 

Have all a great weekend! [preferably less rainy and foggy than here]

António


jtoar:
@AntonioMeireles it's actually pretty foggy out here in LA today; we'll see if that turns into rain

PR #724, opened 2021-06-18
Update Cells doc

jtoar:


thedavidprice:
@jtoar nudge nudge

jtoar:
I'll keep updating this in a separate PR but these updates are good as they are now.

PR #726, opened 2021-06-19
Add notes on purging unused Tailwind CSS

georgewritescode:
Adding notes on purging unused Tailwind CSS to the docs.
It's likely most people will want to purge unused Tailwind CSS when building for production as Tailwind is 3.6mb uncompressed (https://tailwindcss.com/docs/optimizing-for-production). 
For discussion: adding this to the default tailwind config

### DIFF
--- a/docs/webpackConfiguration.md
+++ b/docs/webpackConfiguration.md
@@ -231,13 +231,25 @@ yarn tailwindcss init
 
 This generates the tailwind config file, `tailwind.config.js`. Note that we'd ordinarily move this file to `web/config`, like we did with `postcss.config.js`. But since the Tailwind CSS IntelliSense won't work unless `tailwind.config.js` is in a base directory like `web`, we're just leaving this one here for now.
 
-Finally, use the tailwind directives in `web/src/index.css`
+Next, use the tailwind directives in `web/src/index.css`
 
 ```css
 /* ./web/src/index.css */
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
+```
+
+Optionally, you may also add file types to `purge` in your tailwind config file. This will remove unused tailwind CSS when building for production, reducing your css bundle size.
+```javascript
+// ./web/tailwind.config.js
+  purge: [
+    './src/**/*.html',
+    './src/**/*.jsx',
+    './src/**/*.js',
+    './src/**/*.ts',
+    './src/**/*.tsx',
+  ]
 ```
 
 And that should be it!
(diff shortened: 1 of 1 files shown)


dthyresson:
FYI, in my `tailwind.config.js`, I enable purge only when deploying or in production:

```
  purge: {
    enabled: process.env.NETLIFY || process.env.NODE_ENV === 'production',
    content: [
      './src/**/*.html',
      './src/**/*.js',
      './src/**/*.jsx',
      './src/**/*.tsx',
      './src/**/*.ts',
    ],
  },
```

georgewritescode:
Per the docs here: https://tailwindcss.com/docs/optimizing-for-production

I think it should only run when node env = production
```
Now whenever you compile your CSS with NODE_ENV set to production, Tailwind will automatically purge unused styles from your CSS.

```

However, if Netlify doesn't use node env production, maybe we should add it to the docs

dthyresson:
> However, if Netlify doesn't use node env production, maybe we should add it to the docs

Actually, Netlify does **not* use NODE_ENV production when building -- when building it is `development` which if you think about it is the same as when you build locally because you need all the build and development dependencies.

And, on the web side side it isn't serving -- just hosting static content on a CDN, there is no "production" there either.

See: https://docs.netlify.com/configure-builds/manage-dependencies/#node-js-environment

> Node.js environment

> By default, Netlify’s buildbot sets NODE_ENV to development. You can change this value by setting a NODE_ENV environment variable.

> If you set the NODE_ENV to production, any devDependencies in your package.json file will not be installed for the build.

So, this is why one needs to use `enabled` and look for the `process.env.NODE_ENV === 'production'` (if your hosting doe set that) or `process.env.NETLIFY` because:

> NETLIFY: Always true. Can be used to check if the build is running on Netlify.

See: https://docs.netlify.com/configure-builds/environment-variables/#read-only-variables

So, probably a good idea to add to the docs ;)

georgewritescode:
@dthyresson 
Thanks! I've changed the pull request

thedavidprice:
Thanks for this @georgewritescode 🚀

@dthyresson could you review and merge when ready?

thedavidprice:
@dthyresson gentle nudge. this one looks ready when you are!

Thanks @georgewritescode!

thedavidprice:
@georgewritescode The next Redwood release upgrades Webpack, which allows us to upgrade Tailwind dependencies. And in this PR by @jtoar we are improving the default config to use jit and purge https://github.com/redwoodjs/redwood/pull/3087

How will all the affect this PR?

jtoar:
Thanks @georgewritescode, sorry it took me some time to get to this one. These docs as a whole need updated now that we support tailwind v2, but I'm more than happy to merge this and update them altogether afterwards. Purge is super important to configure—this set up was definitely incomplete without it so thanks again!

PR #727, opened 2021-06-20
add example useMutation hook

ajoslin103:
I added an example of how to use the useMutation hook to the docs-graphql page

### DIFF
--- a/docs/graphql.md
+++ b/docs/graphql.md
@@ -1,4 +1,4 @@
-# GraphQL 
+# GraphQL
 
 GraphQL is a fundamental part of Redwood. Having said that, you can get going without knowing anything about it, and can actually get quite far without ever having to read [the docs](https://graphql.org/learn/). But to master Redwood, you'll need to have more than just a vague notion of what GraphQL is; you'll have to really grok it.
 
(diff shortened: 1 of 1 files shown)


docs/graphql.md
```diff
@@ -28,34 +28,86 @@ const App = () => (
...
+        hasFlag: hasIt,
+      },
+    })
+    .then((data) => {
+      return data
+    })
+}
+```
+
+#### other Appolo hookss
```
dthyresson, line 103: Please correct the typos here.

docs/graphql.md
```diff
@@ -28,34 +28,86 @@ const App = () => (
...
+```
+
+```js
+// api/src/services/users/users.ts
+
+import { db } from 'src/lib/db'
+import { context } from '@redwoodjs/api'
+import { requireAuth } from 'src/lib/auth'
+
+export const setHasPlanner = ({ id = context.currentUser.sub, hasIt }) => {
```
dthyresson, line 88: Typically, services use `async` and `await` with the Prisma `db` client rather than update then return.

dthyresson:
@jtoar Mind giving this a review given your prior documentation work on GraphQL and Cells? Thanks!

docs/graphql.md
ajoslin103: Please see my comment here: https://community.redwoodjs.com/t/updated-i-desire-mutant-cells-solved/2223/10?u=ajoslin103

I don't do this the same way anymore & would like to wait and update this code based on a better understanding and pattern.

I'd like to try my hand at adding MUTATION to the current Cell HOC

ajoslin103:
> @jtoar Mind giving this a review given your prior documentation work on GraphQL and Cells? Thanks!

@jtoar - you could hold off on this review pending my attempt at adding MUTATION to the Cell HOC https://github.com/redwoodjs/redwoodjs.com/pull/727#discussion_r655343540

docs/graphql.md
dthyresson: Ok, then I'll close out this issue.

PR #737, opened 2021-07-01
Clarify that the cli is ran locally in an app and not installed globally

webstackdev:
Just a doc fix, why the CLI can't be installed globally (or where it's installed from) was asked in chat by someone new to Redwood


dthyresson:
Thanks @webstackdev for this PR.

I can definitely see how people may thing RW is setup via the cli vs used within a project.

This helps to clarify.

LGTM. @thedavidprice any thoughts?

thedavidprice:
Yep, definitely helps. Thanks @webstackdev 🚀

PR #743, opened 2021-07-09
Document GraphQL Handler configuration, logger options, and security measures

### LINKED ISSUE #738: Docs: Need to document the GraphQL handler options to configure
With some upcoming enhancements to the GraphQL handler, will want to document the `GraphQLHandlerOptions` available:

```js
interface GraphQLHandlerOptions {
  /**
   * Customize GraphQL Logger
   */
  logger: BaseLogger

  /**
   * Modify the resolver and global context.
   */
  context?: Context | ContextFunction

  /**
   * An async function that maps the auth token retrieved from the request headers to an object.
   * Is it executed when the `auth-provider` contains one of the supported providers.
   */
  getCurrentUser?: GetCurrentUser

  /**
   * A callback when an unhandled exception occurs. Use this to disconnect your prisma instance.
   */
  onException?: () => void

  /**
   * The GraphQL Schema
   */
  schema: GraphQLSchema

  /**
   * CORS configuration
   */
  cors?: CorsConfig

  /**
   * Healthcheck
   */
  onHealthCheck?: OnHealthcheckFn

  /**
   * Limit the complexity of the queries solely by their depth.
   * @see https://www.npmjs.com/package/graphql-depth-limit#documentation
   */
  depthLimit?: DepthLimitConfig

  /**
   * Only allows the specified operation types (e.g. subscription, query or mutation).
   *
   * By default, only allow query and mutation (ie, do not allow subscriptions).
   *
   * @see https://github.com/dotansimha/envelop/tree/main/packages/plugins/filter-operation-type
   */

  allowedOperations?: AllowedOperations

  /**
   * Custom Envelop plugins
   */
  extraPlugins?: Plugin<any>[]

  /**
   * Customize the GraphiQL Endpoint that appears in the location bar of the GraphQL Playground
   *
   * Defaults to '/graphql' as this value must match the name of the `graphql` function on the api-side.
   *
   */
  graphiQLEndpoint?: string

  /**
   * Collect resolver timings, and exposes trace data for
   * an individual request under extensions as part of the GraphQL response.
   */
  tracing?: boolean
}
```

This will apply to ApolloServer as well as has many of the same options.

And also include things like custom `userContext` handlers.

dthyresson:
The PR documents the new options available when configuring the graphql server.

These settings -- including the GraphQL Logger Options - will be the same for both Apollo Server (current) and Helix/Envelop (future).

However, some log examples will need to be updated as the format of some query and trace timing output will be different when using envelop.

This Pr also documents some consideration in place to help secure the graphQL endpoint in development environments by disabling playground an introspection.

It also aims to inform the developer some benefits to using the GraphQL logger -- like keeping your services concise.

I did not go into plugin configuration or other Apollo Server graphql handler setup -- since this is best when transition to envelop and can do a more detailed discussion of the plugin ecosystem/

Closes #738

### DIFF
--- a/docs/graphql.md
+++ b/docs/graphql.md
@@ -371,25 +371,10 @@ Here is an example of an application `/api/src/lib/logger.ts` configured to reda
 
 ```js
 // /api/src/lib/logger.ts
-import { createLogger } from '@redwoodjs/api/logger'
-import { redactionsList } from '@redwoodjs/api/logger'
-/**
- * Creates a logger with RedwoodLoggerOptions
- *
- * These extend and override default LoggerOptions,
- * can define a destination like a file or other supported pin log transport stream,
- * and sets where or not to show the logger configuration settings (defaults to false)
- *
- * @param RedwoodLoggerOptions
- *
- * RedwoodLoggerOptions have
- * @param {options} LoggerOptions - defines how to log, such as pretty printing, redaction, and format
- * @param {string | DestinationStream} destination - defines where to log, such as a transport stream or file
- * @param {boolean} showConfig - whether to display logger configuration on initialization
- */
+import { createLogger, redactionsList } from '@redwoodjs/api/logger'
+
 export const logger = createLogger({
   options: {
-    prettyPrint: true,
     redact: [...redactionsList, 'email', 'data.users[*].email'],
   },
 })
@@ -401,14 +386,13 @@ Often you want to measure and report how long your queries take to execute and r
 
 You may turn on logging these metrics via the `tracing` GraphQL configuration option.
 
+```js
+// api/src/functions/graphql.ts
+// ...
+export const handler = createGraphQLHandler({
+  loggerConfig: { logger, options: { tracing: true } },
+// ...
 ```
-  /**
-   * @description Include the tracing and timing information.
-   *
-   * This will log various performance timings withing the GraphQL event lifecycle (parsing, validating, executing, etc).
-   */
-  tracing?: boolean
-  ```
 
 Let's say we wanted to get some benchmark numbers for the "find post by id" resolver
 
@@ -480,7 +464,7 @@ The [GraphQL Playground](https://github.com/graphql/graphql-playground) is a way
 
 Attackers often submit expensive, nested queries that could overload your database or expend costly resources.
 
-Typically, these types of  complex and expensive queries are usually huge deeply nested and take advantage of an understanding of your schema (hence why schema introspection is disabled byu default in production) and the data model relationships to create "cyclical" queries.
+Typically, these types of  complex and expensive queries are usually huge deeply nested and take advantage of an understanding of your schema (hence why schema introspection is disabled by default in production) and the data model relationships to create "cyclical" queries.
 
 Such unbounded GraphQL queries allow attackers to abuse query depth and with enough depth, this can easily impact your Graphql server and application.
 
@@ -514,15 +498,7 @@ query cyclical {
 
 > To mitigate the risk of attacking your application via deeply nested queries, RedwoodJS by default sets the [Query Depth Limit](https://www.npmjs.com/package/graphql-depth-limit#documentation) to 11. 
 
-If You would like to set the limit to a lower or higher limit, you may do so via the `depthLimitOptions` setting when creating your GraphQL handler.
-
-```
-  /**
-   * Limit the complexity of the queries solely by their depth.
-   * @see https://www.npmjs.com/package/graphql-depth-limit#documentation
-   */
-  depthLimitOptions?: DepthLimitOptions
-```
+You can change the default value via the `depthLimitOptions` setting when creating your GraphQL handler.
 
 You `depthLimitOptions` are `maxDepth` or `ignore` stops recursive depth checking based on a field name. Ignore can be [either a string or regexp]( https://www.npmjs.com/package/graphql-depth-limit#documentation) to match the name, or a function that returns a boolean.
 
--- a/docs/graphql.md
+++ b/docs/graphql.md
@@ -221,6 +221,392 @@ The GraphQL Playground's nice, but if you're a power user, you'll want to be usi
 - dt has some thoughts on this
 - insomnia -->
 
+## Logging
+
+Logging is essential in production apps to be alerted about critical errors and to be able to respond effectively to support issues. In staging and development environments, logging helps you debug queries, resolvers and cell requests.
+
+We want to make logging simple when using RedwoodJS and therefore have configured the api-side GraphQL handler to log common information about your queries and mutations. Log statements also be optionally enriched with [operation names](https://graphql.org/learn/queries/#operation-name), user agents, request ids, and performance timings to give you move visibility into your GraphQL api.
+
+By configuring the GraphQL handler to use your api side [RedwoodJS logger](https://redwoodjs.com/docs/logger), any errors and other log statements about the [GraphQL execution](https://graphql.org/learn/execution/) will be logged to the [destination](https://redwoodjs.com/docs/logger#destination-aka-where-to-log) you've set up: to standard output, file, or transport stream.
+
+You configure the logger using the `loggerConfig` that accepts a [`logger`]((https://redwoodjs.com/docs/logger)) and s set of [GraphQL Logger Options](#graphql-logger-options):
+
+```js
+  loggerConfig: { logger, options: {} }
+```
+
+### Configure the GraphQL Logger
+
+A typical GraphQLHandler `graphql.ts` is as follows:
+
+```js
+// api/src/functions/graphql.ts
+
+import {
+  createGraphQLHandler,
+  makeMergedSchema,
+  makeServices,
+} from '@redwoodjs/api'
+
+import schemas from 'src/graphql/**/*.{js,ts}'
+import { db } from 'src/lib/db'
+import { logger } from 'src/lib/logger'
+import services from 'src/services/**/*.{js,ts}'
+
+export const handler = createGraphQLHandler({
+  loggerConfig: { logger, options: {} },
+  schema: makeMergedSchema({
+    schemas,
+    services: makeServices({ services }),
+  }),
+  onException: () => {
+    // Disconnect from your database with an unhandled exception.
+    db.$disconnect()
+  },
+})
+```
+
+#### Log Common Information
+
+The `loggerConfig` takes several options that logs meaningful information along the graphQL execution lifecycle.
+
+```js
+/**
+ * Options for request and response information to include in the log statements
+ * output by UseRedwoodLogger around the execution event
+ *
+ * @param data - Include response data sent to client.
+ * @param operationName - Include operation name.
+ * @param requestId - Include the event's requestId, or if none, generate a uuid as an identifier.
+ * @param query - Include the query. This is the query or mutation (with fields) made in the request.
+ * @param tracing - Include the tracing and timing information.
+ * @param userAgent -Include the browser (or client's) user agent.
+ */
+type GraphQLLoggerOptions = {
+  /**
+   * @description Include response data sent to client.
+   */
+  data?: boolean
+
+  /**
+   * @description Include operation name.
+   *
+   * The operation name is a meaningful and explicit name for your operation. It is only required in multi-operation documents,
+   * but its use is encouraged because it is very helpful for debugging and server-side logging.
+   * When something goes wrong (you see errors either in your network logs, or in the logs of your GraphQL server)
+   * it is easier to identify a query in your codebase by name instead of trying to decipher the contents.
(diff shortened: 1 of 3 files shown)


docs/graphql.md
```diff
@@ -221,6 +221,405 @@ The GraphQL Playground's nice, but if you're a power user, you'll want to be usi
...
+
+You configure the logger using the `loggerConfig` that accepts a [`logger`]((https://redwoodjs.com/docs/logger)) and s set of [GraphQL Logger Options](#graphql-logger-options):
+
+```js
+  loggerConfig: { logger, options: {} }
+```
+
+### Configure the GraphQL Logger
+
+A typical GraphQLHandler `graphql.ts` is as follows:
```
dac09, line 240: This'll go out of date very quickly I think... maybe just have a smaller snippet passing loggerConfig with some options to createGraphQlHandler?

docs/graphql.md
```diff
@@ -221,6 +221,405 @@ The GraphQL Playground's nice, but if you're a power user, you'll want to be usi
...
+    db.$disconnect()
+  },
+})
+```
+
+#### Log Common Information
+
+The `loggerConfig` takes several options that logs meaningful information along the graphQL execution lifecycle.
+
+```js
```
dac09, line 273: JSDocs are great for highlights, but terrible for readibility in documentation imo (too verbose). Do you think we  can just add a table here with each of the options and describing them?

dthyresson: Yes, a table is a much better idea. Thanks!

docs/graphql.md
```diff
@@ -221,6 +221,405 @@ The GraphQL Playground's nice, but if you're a power user, you'll want to be usi
...
+// api/src/functions/graphql.ts
+// ...
+export const handler = createGraphQLHandler({
+  loggerConfig: { logger, options: { data: true, operationName: true, query: true } },
+// ...
+```
+
+then you do not need to log your `input` variables or the results in each service method.
```
dac09, line 389: For me this whole section on what you should not do doesn't need to be here :)

docs/graphql.md
dthyresson: Makes sense. I take that approach in a few places later in the doc -- just that option itself.

docs/graphql.md
```diff
@@ -221,6 +221,405 @@ The GraphQL Playground's nice, but if you're a power user, you'll want to be usi
...
+  })
+```
+
+We see that this request took about 500 msecs (note: duration is reported in nanoseconds).
+
+Ror more details about the information logged and its format, see [Apollo Tracing](https://github.com/apollographql/apollo-tracing).
+
+```terminal
+pi | INFO [2021-07-09 14:25:52.452 +0000] (apollo-graphql-server): GraphQL willSendResponse
+api |     tracing: {
```
dac09, line 534: Could we shotten this output just to include the important bits? Too much stuff in the documentation feels very daunting

docs/graphql.md
```diff
@@ -230,6 +629,91 @@ We'll document more GraphQL security best practices as Redwood reaches a `v1.0`
...
+   * @see https://www.npmjs.com/package/graphql-depth-limit#documentation
+   */
+  depthLimitOptions?: DepthLimitOptions
+```
+
+You `depthLimitOptions` are `maxDepth` or `ignore` stops recursive depth checking based on a field name. Ignore can be [either a string or regexp]( https://www.npmjs.com/package/graphql-depth-limit#documentation) to match the name, or a function that returns a boolean.
+
+For example:
+
+```js
```
dac09, line 692: Again here... I think we could include just the logger config and take out the rest of the irrelevant code

docs/graphql.md
dthyresson: Perhaps I can rephrase as a suggestion or recommendation.

I have seen on more than one occasion people in the Community take this approach to debugging services and query output:

```js
export const user = () => {
  if (context?.currentUser?.sub) {
    return db.user
      .findUnique({
        where: { id: context?.currentUser?.sub  }
      })
      .then(thenDebug('user'))

  } else {
    return Promise.resolve(null)
  }
};
```

and then

```js
const stringify = require('json-stringify-safe');

import { logger } from './logger'

export const thenDebug = (label) => (data) => {
  logger.debug(`${label}: ${stringify(data, null, 2)}`)
  return data
}
```

And I want to show that it's note necessary to have all that clutter in *every* service and still get the same debugging log info that helps.

That's my intention ... now how best to communicate that. Hm.

docs/graphql.md
```diff
@@ -221,6 +221,405 @@ The GraphQL Playground's nice, but if you're a power user, you'll want to be usi
...
+
+but keep your services concise!
+
+#### Send to Third-party Transports
+
+Stream to third-party log and application monitoring services vital to production logging in serverless environments like [logFlare](https://logflare.app/), [Datadog](https://www.datadoghq.com/) or [LogDNA](https://www.logdna.com/)
+
+#### Supports Log Redaction
+
+Everyone has herd or reports that Company X logged emails, or passwords to files or systems that may not have been secured. While RedwoodJS logging won't necessarily prevent that, it does provide you with the mechanism to ensure that won't happen.
```
dac09, line 469: herd -> heard

docs/graphql.md
dthyresson: :+1: I'll "ellipses" out several of the resolver path timings ... the top level duration is the most important.

docs/graphql.md
dac09: Yeah, the problem with including "dont do this" snippets is that its very easy to miss the "dont" part.

Just adding a sentence at the end to say this will log whatever you need, without needing to add additional logger statements into your services

docs/graphql.md
```diff
@@ -221,6 +221,405 @@ The GraphQL Playground's nice, but if you're a power user, you'll want to be usi
...
+
+Therefore, if you wish to log the GraphQL `query` made, the `data` returned, and the `operationName` used, you would
+
+```js
+// api/src/functions/graphql.ts
+
+export const handler = createGraphQLHandler({
+  loggerConfig: {
+    logger,
+    options: { data: true, operationName: true, query: true },
```
dac09, line 340: Using a diff here and just showing the 3/4 relevant lines here would probably make it more digestable.

docs/graphql.md
```diff
@@ -221,6 +221,405 @@ The GraphQL Playground's nice, but if you're a power user, you'll want to be usi
...
+})
+```
+
+#### Timing Traces for Benchmarking Performances
+
+Often you want to measure and report how long your queries take to execute and respond. You may already be measuring these durations at the database level, but you can also measure the time it takes for your the GraphQL server to parse, validate, and execute the request.
+
+You may turn on logging these metrics via the `tracing` GraphQL configuration option.
+
+```
```
dac09, line 511: Too much repeated info here I think. Just showing 
```
createGraphQlHandler({
	loggerOptions: {logger: {tracing: true}}
	/*
	..other options
	*/
```

Probably more helpful

docs/graphql.md
```diff
@@ -221,6 +221,405 @@ The GraphQL Playground's nice, but if you're a power user, you'll want to be usi
...
+Because this logger is used with the GraphQL handler, it will respect any redaction paths setup.
+
+For example, you have chosen to log `data` return by each request, then you may want to redact sensitive information, like email addresses from yur logs.
+
+Here is an example of an application `/api/src/lib/logger.ts` confgured to redact email addresses. Take note of the path `data.users[*].email` as this says, in the `data` attribute, redact the `email` from every `user`:
+
+```js
+// /api/src/lib/logger.ts
+import { createLogger } from '@redwoodjs/api/logger'
+import { redactionsList } from '@redwoodjs/api/logger'
```
dac09, line 482: Suggested snippet: 
```diff
// /api/src/lib/logger.ts
+import { createLogger, redactionsList } from '@redwoodjs/api/logger'

export const logger = createLogger({
  options: {
+    redact: [...redactionsList, 'email', 'data.users[*].email'],
  },
})
```

docs/graphql.md
```diff
@@ -221,6 +221,405 @@ The GraphQL Playground's nice, but if you're a power user, you'll want to be usi
...
+
+Logging is essential in production apps to be alerted about critical errors and to be able to respond effectively to support issues. In staging and development environments, logging helps you debug queries, resolvers and cell requests.
+
+We want to make logging simple when using RedwoodJS and therefore have configured the api-side GraphQL handler to log common information about your queries and mutations. Log statements also be optionally enriched with [operation names](https://graphql.org/learn/queries/#operation-name), user agents, request ids, and performance timings to give you move visibility into your GraphQL api.
+
+By configuring the GraphQL handler to use your api side [RedwoodJS logger](https://redwoodjs.com/docs/logger), any errors and other log statements about the [GraphQL execution](https://graphql.org/learn/execution/) will be logged to the [destination](https://redwoodjs.com/docs/logger#destination-aka-where-to-log) you've set up: to standard output, file, or transport stream.
+
+You configure the logger using the `loggerConfig` that accepts a [`logger`]((https://redwoodjs.com/docs/logger)) and s set of [GraphQL Logger Options](#graphql-logger-options):
```
dac09, line 232: Think 232-236 is repeated from the next section

docs/graphql.md
```diff
@@ -230,6 +629,91 @@ We'll document more GraphQL security best practices as Redwood reaches a `v1.0`
...
+
+> Because both introspection and the playground share possibly sensitive information about your data model, your data, your queries and mutations, best practices for deploying a GraphQL Server call to disable these in production, RedwoodJS **only enables introspection and the playground when running in development**. That is when `process.env.NODE_ENV === 'development'`.
+
+### Query Depth Limit
+
+Attackers often submit expensive, nested queries that could overload your database or expend costly resources.
+
+Typically, these types of  complex and expensive queries are usually huge deeply nested and take advantage of an understanding of your schema (hence why schema introspection is disabled byu default in production) and the data model relationships to create "cyclical" queries.
+
+Such unbounded GraphQL queries allow attackers to abuse query depth and with enough depth, this can easily impact your Graphql server and application.
```
dac09, line 646:
```suggestion
```

docs/graphql.md
```diff
@@ -230,6 +629,91 @@ We'll document more GraphQL security best practices as Redwood reaches a `v1.0`
...
+        }
+      }
+    }
+  }
+}
+```
+
+> To mitigate the risk of attacking your application via deeply nested queries, RedwoodJS by default sets the [Query Depth Limit](https://www.npmjs.com/package/graphql-depth-limit#documentation) to 11. 
+
+If You would like to set the limit to a lower or higher limit, you may do so via the `depthLimitOptions` setting when creating your GraphQL handler.
```
dac09, line 556:
```suggestion
You can change the default value via the `depthLimitOptions` setting when creating your GraphQL handler.
```

docs/graphql.md
```diff
@@ -230,6 +629,91 @@ We'll document more GraphQL security best practices as Redwood reaches a `v1.0`
...
+
+```
```
dac09, line 558: We can take this jsdoc out I think

docs/graphql.md
dthyresson: Got it. Yes, I can see how by showing example of what not to do, people might think of it as something to do (by accident).

I made some revisions.

docs/graphql.md
dthyresson: Agreed. Thanks.

docs/graphql.md
dthyresson: Agreed

docs/graphql.md
dthyresson: Too scary? It's true though.

docs/graphql.md
dthyresson: Yup

docs/graphql.md
dthyresson: Agreed

docs/graphql.md
dthyresson: Sorry, I don't follow. Can you explain given the new updated document? Thanks!

docs/graphql.md
dac09: Nah I think you've mentioned this info in the above and below paragraphs

docs/graphql.md
dac09: I just mean line 232-236 is repeated content, because you're showing how to configure a logger in line 249.

docs/graphql.md
dthyresson: Combined the first sentence and this one to use some same terms, but shorten this section.

docs/graphql.md
dthyresson: Got it. removing.

dac09 (approved):
LGTM! 👏

PR #745, opened 2021-07-11
Updates to the Data Migrate docs

cjreimer:
This PR updates a couple of items in the data migrate docs to the current status.

### DIFF
--- a/docs/dataMigrations.md
+++ b/docs/dataMigrations.md
@@ -151,7 +151,7 @@ export default async ({ db }) => {
 
 Run once:
 
-    yarn rw dataMigrate install
+    yarn rw data-migrate install
     yarn rw prisma migrate dev
 
 Run every time you need a new data migration:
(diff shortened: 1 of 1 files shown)


dac09 (approved):

docs/dataMigrations.md
```diff
@@ -152,7 +152,7 @@ export default async ({ db }) => {
 Run once:
 
     yarn rw dataMigrate install
```
dac09, line 154:
```suggestion
    yarn rw data-migrate install
```
We're trying to use the kebab case version in the cli commands (although camelCase is still supported). Ok to add this change in?

cjreimer:
Ok, with @dac09 's suggestion, and I found https://github.com/redwoodjs/redwood/pull/2988, I've changed the other dataMigration references to data-migration.

Note that 
`yarn rw generate dataMigration migrationName`

appears to be still current in the codebase, which is a bit inconsistent.

This pr now should only be merged when https://github.com/redwoodjs/redwood/pull/2988 is merged

PR #748, opened 2021-07-15
Documents how to customize and enrich the context in the GraphQL Handler

dthyresson:
As part of the GraphQL Helix / Envelop testing, it happens that customizing and populating the context in the GraphQL handler on a per-request basis using a custom ContextFunction changed.

In the api/Apollo version, the ContextFunction receives both the event and the context.

In envelop, a plugin has the just context, but the event information has been placed on the context already.

Therefore, if you had a function that used the event, one just now get the event from `context.event` when migrating to use `graphql-server`.

In order to describe this change in the Release Notes / Breaking Change / How to Upgrade section, there needs to be some existing documentation about how to enrich the context -- and this PR includes this info.

### DIFF
--- a/docs/graphql.md
+++ b/docs/graphql.md
@@ -195,7 +195,7 @@ import { context } from '@redwoodjs/api
 
 Because the context is read-only in your services, if you need to modify it, then you need to do so in the `createGraphQLHandler`.
 
-To populate or enrich the context on a per-request basis with additional attributes, set a the `context` attribute `createGraphQLHandler` to a custom ContextFunction which modifies the context.
+To populate or enrich the context on a per-request basis with additional attributes, set the `context` attribute `createGraphQLHandler` to a custom ContextFunction that modifies the context.
 
 For example, if we want to populate a new, custom `ipAddress` attribute on the context with the information from the request's event, declare the `setIpAddress` ContextFunction as seen here:
 
--- a/docs/graphql.md
+++ b/docs/graphql.md
@@ -234,6 +234,8 @@ export const handler = createGraphQLHandler({
 })
 ```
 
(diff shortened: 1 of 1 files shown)


dthyresson:
Once approved, I can then add a link to this documentation in the Release Notes and Breaking Changes when migrating to Helix / Envelop.

docs/graphql.md
```diff
@@ -191,6 +191,49 @@ In Redwood, the `context` object that's passed to resolvers is actually availabl
 import { context } from '@redwoodjs/api
 ```
 
+#### How to Modify the Context
+
+Because the context is read-only in your services, if you need to modify it, then you need to do so in the `createGraphQLHandler`.
+
+To populate or enrich the context on a per-request basis with additional attributes, set a the `context` attribute `createGraphQLHandler` to a custom ContextFunction which modifies the context.
```
thedavidprice, line 198:
```suggestion
To populate or enrich the context on a per-request basis with additional attributes, set the `context` attribute `createGraphQLHandler` to a custom ContextFunction that modifies the context.
```

thedavidprice (approved):
Approved!

@dthyresson I didn't understand if this is relevant _now_ and should be merged, or is this for v0.35 and should be merged once that release is published?

dthyresson:
> is relevant _now_ and should be merged,

It is relevant now in 0.35 with Apollo Server. 

When and if we use Envelop, then these docs change slightly to fetch the event differently.

I guess I should add a note in these docs concerning the differences.

dthyresson:
> add a note in these docs concerning the differences.

@thedavidprice I added I note here about Helix/Envelop.

thedavidprice:
Merged! (Hoping that's what you meant for me to do... 🤞)

PR #758, opened 2021-07-26
Use example app name in TS docs

chenkie:
It might be best to keep consistent with the Quickstart docs and use `my-redwood-app` as a concrete app name. I got a bit confused when reading `/PATH/TO/FOLDER` when getting started with the TS docs :)


dthyresson (approved):
Thanks for keeping our docs consistent. Super helpful especially for people just getting started.

PR #760, opened 2021-07-30
Logger: Added a snippet for Honeybadger

noire-munich:
This PR adds a section in Logger to help anyone logging to Honeybadger.

Pino doesn't have - at least I haven't found any, a transport package for [Honeybadger](https://honeybadger.io), but we can write one with the class `Write` from the `stream` package. That can serve as a modest sample for anyone who'd want to implement a more custom logging solution.

### DIFF
--- a/docs/logger.md
+++ b/docs/logger.md
@@ -431,6 +431,52 @@ export const logger = createLogger({
   destination: stream },
 }) 
 ```
+### Log to Honeybadger using a Transport Stream Destination
+
+* Install the Honeybadger & Stream packages into `api`
+```shell
+yarn workspace api add @honeybadger-io/js stream
+```
+
+* Import `@honeybadger-io/js` `stream` into `logger.ts`
+```js
+import { createLogger } from '@redwoodjs/api/logger'
+import { Writable } from 'stream'
+
+const Honeybadger = require('@honeybadger-io/js')
+
+Honeybadger.configure({
+    apiKey: process.env.HONEYBADGER_API_KEY,
+})
+
+const HoneybadgerStream = () => {
+    const stream = new Writable({
+        write(
+            chunk: any,
+            encoding: BufferEncoding,
+            fnOnFlush: (error?: Error | null) => void
+        ) {
+            Honeybadger.notify(chunk.toString())
+            fnOnFlush()
+        },
+    })
+
+    return stream
+}
+
+
+/**
+ * Creates a logger. Options define how to log. Destination defines where to log.
+ * If no destination, std out.
+ */
+export const logger = createLogger({
+    options: { prettyPrint: true },
+    destination: HoneybadgerStream(),
+})
+```
+
+* Make sure you have a `HONEYBADGER_API_KEY` variable in your environment.
+
 ### Log to Papertrail using a Transport Stream Destination
 
 * Install the [pino-papertrail](https://www.npmjs.com/package/pino-papertrail) package into `api`
--- a/docs/logger.md
+++ b/docs/logger.md
@@ -306,9 +306,8 @@ export const logger = createLogger({
 ```
 
 ### Customize your own Transport Stream Destination ( eg: Honeybadger.io )
-If Pino doesn't have a transport package for your service, you can write one with the class `Write` from the `stream` package. You can adapt this example to your own needs but here, we will use Honeybadger.io, which doesn't have an official pino transport package.
+If `pino` doesn't have a transport package for your service, you can write one with the class `Write` from the `stream` package. You can adapt this example to your own logging needs but here, we will use [Honeybadger.io](https://honeybadger.io).
 
-This example where we log to Honeybadger is an example of how to implement a custom logging solution.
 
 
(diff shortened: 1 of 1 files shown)


dthyresson:
> Pino doesn't have - at least I haven't found any, a transport package for [Honeybadger](https://honeybadger.io), but we can write one with the class `Write` from the `stream` package

Thanks for this PR and excellent example.

In fact, perhaps calling out that one can implement your own logger destination is valuable in the docs?

Something like:

```
If Pino doesn't have a transport package for your service, you can write one with the class Write from the stream package. This example where we log to Honeybadger is an example of how to implement a custom logging solution.
```

Or words to that effect?

noire-munich:
> > Pino doesn't have - at least I haven't found any, a transport package for [Honeybadger](https://honeybadger.io), but we can write one with the class `Write` from the `stream` package
> 
> Thanks for this PR and excellent example.
> 
> In fact, perhaps calling out that one can implement your own logger destination is valuable in the docs?
> 
> Something like:
> 
> ```
> If Pino doesn't have a transport package for your service, you can write one with the class Write from the stream package. This example where we log to Honeybadger is an example of how to implement a custom logging solution.
> ```
> 
> Or words to that effect?

Thanks @dthyresson, it does make sense indeed.

I borrowed your words but twisted them a bit and renamed the example as ```Customize your own Transport Stream Destination ( eg: Honeybadger.io )```, to make the customization more important than the choice of library. Is it ok this way, or should I revert to something more in the line of the previous title and other examples?

dthyresson:
> renamed example as `Customize your own Transport Stream Destination ( eg: Honeybadger.io )`,

Unfortunately, I think the ( parenthesis or maybe : is breaking the table of contents link here (can preview):

https://deploy-preview-760--redwoodjs.netlify.app/docs/logger#configuration-examples

The Datadog logger link examples do visit the anchor, but this one doesn't.

Maybe add the Honeybadger info as a different header?

Or maybe

`Customize your own Transport Stream Destination like Honeybadger.io` ?

noire-munich:
> Unfortunately, I think the ( parenthesis or maybe : is breaking the table of contents link here (can preview):

Neither of them, oddly enough! Attempting some fixes until it works

dthyresson:
> Neither of them, oddly enough! Attempting some fixes until it works

Unfortunately those docs links are really odd and use some for of turbolinks and are really picky about characters and even depth.

Thanks for keeping at it, I know I use that TOC all the time and is the best way to get to docs fast. Appreciate it.

dthyresson:
Maybe "Honeybadger Example of Custom Log Destination" ?

noire-munich:
> Thanks for keeping at it, I know I use that TOC all the time and is the best way to get to docs fast. Appreciate it.

No problem, I do as well and I'm seriously infuriated when browsing a site with a broken TOC x).

noire-munich:
> Maybe "Honeybadger Example of Custom Log Destination" ?

I got `Customize your own Transport Stream Destination, eg: with Honeybadger` working! I think it really didn't like the "honeybadger.io", specially the `.io` x).

So, we shall use whichever you think is best :)

dthyresson (approved):
Great example.

BTW Release 7.0 makes a number of changes in Pino, one of which is to support multiple stream transport destinations.

See: https://github.com/pinojs/pino/releases/tag/v7.0.0-rc.1

So, may need up update those docs when Redwood upgrades.

noire-munich:
> BTW Release 7.0 makes a number of changes in Pino, one of which is to support multiple stream transport destinations.

I did see these indeed! Do you have any estimate for when RW will upgrade to Pino v7?

Thanks a lot for the merge & the help :+1:

dthyresson:
I did a proof as part of https://github.com/redwoodjs/redwood/issues/3090 but some imports were missing or not 100% sure how to change -- ie I couldn't import the BaseLogger.

If you'd like to work on https://github.com/redwoodjs/redwood/issues/3090, please let me know and happy to have you try.

I was going to wait for a on an -rc release.

noire-munich:
> If you'd like to work on [redwoodjs/redwood#3090](https://github.com/redwoodjs/redwood/issues/3090), please let me know and happy to have you try.


I think I may give it a try within our current SportOffice context: we've got troubles on our preprod env and having transports based on the log level could help us make some sense of it. I'll fiddle with the idea and let you know if anything comes out of it^^

dthyresson:
> > If you'd like to work on [redwoodjs/redwood#3090](https://github.com/redwoodjs/redwood/issues/3090), please let me know and happy to have you try.
> 
> I think I may give it a try within our current SportOffice context: we've got troubles on our preprod env and having transports based on the log level could help us make some sense of it. I'll fiddle with the idea and let you know if anything comes out of it^^

Wonderful!

I'd appreciate yours thoughts how one might change the `destination` option to support multiple ones , too.

noire-munich:
> I'd appreciate your thoughts how one might change the `destination` option to support multiple ones, too.

Sure! Replied in https://github.com/redwoodjs/redwood/issues/3090

Issue #735, opened 2021-06-30
Docs: Environment Variables doc lies like a rug when it comes to ENV vars on the web side


cannikin:
[The docs](https://redwoodjs.com/docs/environment-variables#development) say that if you create any environment variables they will be available on the web site with no special config, but it turns out this is not true! @jtoar and I found that the steps that apply to production are also required on in dev: either prefix the ENV var with `REDWOOD_ENV_` or add an option in `redwood.toml` to allow a list of individual ENV vars that should be made available (and then you can name them whatever you want).

We can probably just remove the *Development* section completely and remove the *Production* heading—those instructions apply to the web side, full stop.

PR #767, opened 2021-08-07
makes ref to env vars on web clearer

sarthaktexas:
ref to #735

### DIFF
--- a/docs/environmentVariables.md
+++ b/docs/environmentVariables.md
@@ -17,38 +17,6 @@ Redwood also configures Webpack with `dotenv-webpack`, so that all references to
 
 ## Web
 
-### Development
-
-Like the API Side, anything defined in your `.env` and `.env.defaults` files is simply available:
-
-```
-HELLO_ENV=hello world
-```
-
-```javascript{8}
-// web/src/pages/HomePage.js
-
-const HomePage = () => {
-  return (
-    <div>
-      <h1>HomePage</h1>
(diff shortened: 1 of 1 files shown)


dthyresson:
@cannikin As you authored https://github.com/redwoodjs/redwoodjs.com/issues/735 have assigned for your review.

Though, this no now longer explain how to use a personal, local `.env` or how to use the `.defaults.env` though it is somewhat referred to in the above section (see Web).

I think one of the issues is that the `includeEnvironmentVariables = ` is not set to `[]` in the default create redwood app template -- so it is easy to forget to set them in dev ... unless you refer to these docs.

Maybe that setting is added by default?

thedavidprice:
Quick FYI that @cannikin is taking a vacation this week. He'll be back next week. Please don't hesitate to nudge as needed after Aug 18th if you haven't heard from him.

cannikin (approved):

cannikin:
@dthyresson Good call, should we update CRA to include that in the redwood.toml template? We can add a comment above with a quick explanation as well...

sarthaktexas:
sorry for being dumb, what's CRA? i can totally make a pr to do whatever it is that yall are saying too

aidenybai:
This is very epic of you sarthak.

sarthaktexas:
@cannikin can you make a new issue regarding this?

cannikin:
@sarthaktexas There you go!  👆

Issue #104, opened 2020-04-23
Add some kind of "follow us on twitter" widget to homepage


cannikin:

jtoar:
@cannikin Does our twitter bade in the top right count?

cannikin:
This came out of a random meeting with the core team and we were throwing out ideas. @mojombo mentioned some kind of call out for "Follow us on Twitter", more prominent than the icon at the upper right, so I added this issue, then forgot about it. 😬

sarthaktexas:
Is there a specific reason why the twitter should be more prominent?

sarthaktexas:
that makes sense

PR #768, opened 2021-08-07
adds follow us on twitter as a thing

sarthaktexas:
ref to #104


sarthaktexas:
I closed this because on second thought, I can't think of a way to actually put this on the site in a visually pleasing way.

PR #774, opened 2021-08-11
How to test functions and webhooks

dthyresson:
Describes how to test the serverless functions and webhooks with code examples.

TODO:

- [x] Need to test again with the latest merged framework as had just tested with the open PR.
- [x] prettify code
- [x] code formatting to fit in narrow columns
- [x] feedback from Danny's review
- [x] Add advanced function mocking with fixtures

### DIFF
--- a/docs/serverlessFunctions.md
+++ b/docs/serverlessFunctions.md
@@ -310,7 +310,7 @@ export const handler = async (event: APIGatewayEvent) => {
 }
 ```
 
-##### Webhook Test Scenarios
+#### Webhook Test Scenarios
 
 ```ts
 // api/src/functions/updateOrderStatus/updateOrderStatus.scenarios.ts
--- a/docs/serverlessFunctions.md
+++ b/docs/serverlessFunctions.md
@@ -116,8 +116,7 @@ export const handler = async (event: APIGatewayEvent) => {
   try {
 
     // get the two numbers to divide from the event query string
-    const dividend = event.queryStringParameters.dividend
-    const divisor = event.queryStringParameters.divisor
+    const { dividend, divisor } = event.queryStringParameters.dividend
 
     // make sure the values to divide are provided
     if (dividend === undefined || divisor === undefined) {
--- a/docs/serverlessFunctions.md
+++ b/docs/serverlessFunctions.md
@@ -181,6 +181,7 @@ The setup steps are to:
 * invoke the handler with the mocked event
 * extract the result body
 * test that the values match what you expect
+
 The boilerplate steps are generated automatically for you by the function generator
 Let's look at a series of tests that mock the event with different information in each.
 
@@ -293,7 +294,9 @@ In the following example, we'll have the webhook interact with our app's databas
 
 > **Why testing webhooks is hard**
 >
->Because your webhook is typically sent from a third-party's system, manually testing webhooks can be difficult. For one thing, you often have to create some kind of event in their system that will trigger the event -- and you'll often have to do that in a production environment with real data. Second, for each case you'll have to find data that represents each case and issue a hook for each -- which can take a lot of time and is tedious. Also, you'll be using production secrets to sign the payload. And finally, since your third-party needs to send you the incoming webhook you'll most likely have to launch a local tunnel to expose your development machine publicly in order to receive them.
+>Because your webhook is typically sent from a third-party's system, manually testing webhooks can be difficult. For one thing, you often have to create some kind of event in their system that will trigger the event -- and you'll often have to do that in a production environment with real data. Second, for each case you'll have to find data that represents each case and issue a hook for each -- which can take a lot of time and is tedious. 
+>
+>Also, you'll be using production secrets to sign the payload. And finally, since your third-party needs to send you the incoming webhook you'll most likely have to launch a local tunnel to expose your development machine publicly in order to receive them.
 >
 >Instead, we can automate and mock the webhook to contain a signed payload that we can use to test the handler.
 >
@@ -392,7 +395,8 @@ export const handler = async (event: APIGatewayEvent) => {
 
     // updated the order with the new status using the trackingNumber provided
     const order = await db.order.update({
-      where: { trackingNumber_status: {trackingNumber, status: currentOrderStatus }},
+      where: { trackingNumber_status: { trackingNumber, 
+                                        status: currentOrderStatus }},
       data: { status: status },
     })
 
@@ -468,16 +472,18 @@ import { mockSignedWebhook } from '@redwoodjs/testing/api'
 import { handler } from './updateOrderStatus'
 
 describe('updates an order via a webhook', () => {
-  scenario('with a shipped order, updates the status to DELIVERED', async (scenario) => {
+  scenario('with a shipped order, updates the status to DELIVERED', 
+            async (scenario) => {
+
     const order = scenario.order.shipped
 
     const payload = { trackingNumber: order.trackingNumber,
                       status: 'DELIVERED' }
 
     const event = mockSignedWebhook({ payload,
-                                      signatureType: 'sha256Verifier',
-                                      signatureHeader: 'X-Webhook-Signature',
-                                      secret: 'MY-VOICE-IS-MY-PASSPORT-VERIFY-ME' })
+                      signatureType: 'sha256Verifier',
+                      signatureHeader: 'X-Webhook-Signature',
+                      secret: 'MY-VOICE-IS-MY-PASSPORT-VERIFY-ME' })
 
     const result = await handler(event)
 
@@ -499,13 +505,16 @@ Because the header isn't what the webhook expects (it wants to see a header name
 
 ```javascript
 
-  scenario('with an invalid signature header, the webhook is unauthorized', async (scenario) => {
+  scenario('with an invalid signature header, the webhook is unauthorized', 
+            async (scenario) => {
     const order = scenario.order.placed
 
-    const payload = {trackingNumber: order.trackingNumber, status: 'DELIVERED'}
-    const event = mockSignedWebhook({payload, signatureType: 'sha256Verifier',
-                       signatureHeader: 'X-Webhook-Signature-Invalid',
-                       secret: 'MY-VOICE-IS-MY-PASSPORT-VERIFY-ME' })
+    const payload = { trackingNumber: order.trackingNumber, 
+                      status: 'DELIVERED' }
+    const event = mockSignedWebhook({ payload, 
+                      signatureType: 'sha256Verifier',
+                      signatureHeader: 'X-Webhook-Signature-Invalid',
+                      secret: 'MY-VOICE-IS-MY-PASSPORT-VERIFY-ME' })
 
     const result = await handler(event)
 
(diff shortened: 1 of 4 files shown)


dthyresson:
> because the new generator will do it for us

Ah, yes it will.

docs/serverlessFunctions.md
```diff
@@ -45,6 +37,533 @@ Note that you can use code in `api/src` in your serverless function, such as imp
...
+Test Suites: 1 passed, 1 total
+Tests:       4 passed, 4 total
+Snapshots:   0 total
+Time:        13.155 s
+Ran all test suites matching /divide.test.ts|divide.test.ts|false/i.
+
+Active Filters: filename /divide.test.ts|divide.test.ts|false/
+ › Press c to clear filters.
+
+Watch Usage
```
dac09, line 262: Do you think we could remove line 259-260? And the same in the webhooks one?

docs/serverlessFunctions.md
```diff
@@ -45,6 +37,533 @@ Note that you can use code in `api/src` in your serverless function, such as imp
...
+
+[Webhooks](https://redwoodjs.com/docs/webhooks#webhooks) are specialized serverless functions that will verify a signature header to ensure you can trust the incoming request and use the payload with confidence.
+
+> Note: Want to learn more about webhooks? See a [Detailed discussion of webhooks](https://redwoodjs.com/docs/webhooks) to find out how webhooks can give your app the power to create complex workflows, build one-to-one automation, and sync data between apps.
+
+Because your webhook is typically sent from a third-party's system, manually testing webhooks can be difficult. For one thing, you often have to create some kind of event in their system that will trigger the event -- and you'll often have to do that in a production environment with real data. Second, for each case you'll have to find data that represents each case and issue a hook for each -- which can take a lot of time and is tedious. Also, you'll be using production secrets to sign the payload. And finally, since your third-party needs to send you the incoming webhook you'll most likely have to launch a local tunnel to expose your development machine publicly in order to receive them.
```
dac09, line 280: Maybe put this in a highlight block? 
> **Why testing webhooks is hard**
> Because your webhook is typically sent from a third-party's system, manually testing webhooks can be difficult. For one thing, you often have to create some kind of event in their system that will trigger the event -- and you'll often have to do that in a production environment with real data. Second, for each case you'll have to find data that represents each case and issue a hook for each -- which can take a lot of time and is tedious. Also, you'll be using production secrets to sign the payload. And finally, since your third-party needs to send you the incoming webhook you'll most likely have to launch a local tunnel to expose your development machine publicly in order to receive them. 
	

> Instead, we can automate and mock the webhook to contain a signed payload that we can use to test the handler.

docs/serverlessFunctions.md
dthyresson: Yup. I meant to but forgot.

docs/serverlessFunctions.md
dthyresson: That's better. Yes.

docs/serverlessFunctions.md
```diff
@@ -45,6 +37,533 @@ Note that you can use code in `api/src` in your serverless function, such as imp
...
+
+Because your webhook is typically sent from a third-party's system, manually testing webhooks can be difficult. For one thing, you often have to create some kind of event in their system that will trigger the event -- and you'll often have to do that in a production environment with real data. Second, for each case you'll have to find data that represents each case and issue a hook for each -- which can take a lot of time and is tedious. Also, you'll be using production secrets to sign the payload. And finally, since your third-party needs to send you the incoming webhook you'll most likely have to launch a local tunnel to expose your development machine publicly in order to receive them. 
+
+Instead, we can automate and mock the webhook to contain a signed payload that we can use to test the handler.
+
+By writing these tests, you cam iterate and work on implementing the webhook logic much faster and easier because you don't have to rely on a third party to send you data, or having to setup tunnelling, or triggering events on the external system or service.
```
dac09, line 284:
```suggestion
By writing these tests, you can iterate and implement the webhook logic much faster and easier without having to rely on a third party to send you data, or setting up tunnelling, or triggering events on the external system or service.
```

docs/serverlessFunctions.md
```diff
@@ -45,6 +37,533 @@ Note that you can use code in `api/src` in your serverless function, such as imp
...
+
+In the following example, we'll also have the webhook interact with our app's database, so we can see how we can use scenario testing to create data that the handler can access and modify.
```
dac09, line 286:
```suggestion
In the following example, we'll also have the webhook interact with our app's database, so we can see how we can use **scenario testing** to create data that the handler can access and modify.
```

docs/serverlessFunctions.md
```diff
@@ -45,6 +37,533 @@ Note that you can use code in `api/src` in your serverless function, such as imp
...
+  createdAt      DateTime @default(now())
+  updatedAt      DateTime @updatedAt
+  trackingNumber String   @unique
+  status         String   @default("UNKNOWN")
+
+  @@unique([trackingNumber, status])
+}
+```
+
+Our function is called `updateOrderStatus` and exists in the `api/src/functions` directory along with the tests and scenario scripts:
```
dac09, line 306: Maybe replace this section with - 

Let's generate our webhook function:
```
yarn rw generate function updateOrderStatus
```

docs/serverlessFunctions.md
```diff
@@ -45,6 +37,533 @@ Note that you can use code in `api/src` in your serverless function, such as imp
...
+    }
+  }
+}
+```
+
+#### Webhook Test Scenarios
+
+Since our `updateOrderStatus` webhook will query an order by its tracking number and then attempt to update its status, we'll want to seed our test run with some scenario data that helps us have records we can use to test that the webhook does what we expect it to in each situation.
+
+Let's create three orders for with different status: places, shipped, and delivered.
```
dac09, line 409:
```suggestion
Let's create three orders for with different status: `places`, `shipped`, and `delivered`.
```

docs/serverlessFunctions.md
```diff
@@ -102,9 +621,51 @@ export const handler = async (event, context) => {
 }
 ```
 
+#### Running Webhook Tests
+
+To run an individual webhook test:
+
+```terminal
+// cd into `api/src/functions/updateOrderStatus`
+yarn rw test updateOrderStatus.test.ts
```
dac09, line 630:
```suggestion
yarn rw test api updateOrderStatus
```

docs/serverlessFunctions.md
```diff
@@ -102,9 +621,51 @@ export const handler = async (event, context) => {
 }
 ```
 
+#### Running Webhook Tests
+
+To run an individual webhook test:
+
+```terminal
+// cd into `api/src/functions/updateOrderStatus`
```
dac09, line 629:
```suggestion
```

docs/serverlessFunctions.md
```diff
@@ -102,9 +621,51 @@ export const handler = async (event, context) => {
 }
 ```
 
+#### Running Webhook Tests
```
dac09, line 624: I think you meant to have this above the security section?

docs/serverlessFunctions.md
```diff
@@ -45,6 +37,533 @@ Note that you can use code in `api/src` in your serverless function, such as imp
...
+
+```js
+// /api/db/schema.prisma
+
+model Order {
+  id             Int      @id @default(autoincrement())
+  createdAt      DateTime @default(now())
+  updatedAt      DateTime @updatedAt
+  trackingNumber String   @unique
+  status         String   @default("UNKNOWN")
```
dac09, line 300: No need to change this right now, we can pair on it tomorrow if need be, but what do you think of changing this to an Enum? Here's an example from the tape codebase

```js
//prisma schema
model TapePlan {
  id               String             @id @default(cuid())
  //.
  //.
  TapeSubscription TapeSubscription[]
}

enum SubscriptionStatus {
  ACTIVE
  DORMANT
  TRIALING
  FAILED_PAYMENT
}
```

And you can use it like this:

```js
import { SubscriptionStatus } from '@prisma/client'

/*
.
. */

if (
updatedSubscription.status === SubscriptionStatus.DORMANT
) {
// do some stuff
}

```

dthyresson: I was going to use an Enum, but they are not supported by SQLite.

image

dthyresson: Which is a shame, because it fits so well here. Maybe I'll make a note.

dac09: Oh gotcha... good point.

dac09:
Oh also... we need to add a link from the testing docs to the functions testing section too!

dthyresson:
> . we need to add a link from the testing docs to the functions testing section too!

Done!

docs/serverlessFunctions.md
```diff
@@ -19,92 +11,622 @@ Creating serverless functions is easy with Redwood's function generator:
...
+│   │   ├── divide
+│   │   │   ├── divide.ts
+│   │   │   ├── divide.test.ts
+```
 
-When you're developing locally, the dev server watches the `api` directory for modifications; when it detects any, it re-imports all the modules.
+The setup steps are to:
 
-## Security considerations
+* import your api testing utilities, such as `mockHttpEvent`
```
dac09, line 180:
```suggestion
```

docs/serverlessFunctions.md
```diff
@@ -19,92 +11,622 @@ Creating serverless functions is easy with Redwood's function generator:
...
+* import your function handler
```
dac09, line 181:
```suggestion
```

docs/serverlessFunctions.md
```diff
@@ -19,92 +11,622 @@ Creating serverless functions is easy with Redwood's function generator:
...
+* write your test cases by mocking the event to contain the information you want to give the handler
```
dac09, line 182:
```suggestion
* write your test cases by mocking the event using `mockHttpEvent` to contain the information you want to give the handler
```

docs/serverlessFunctions.md
```diff
@@ -19,92 +11,622 @@ Creating serverless functions is easy with Redwood's function generator:
...
+* invoke the handler with the mocked event
+* extract the result body
+* test that the values match what you expect
```
dac09, line 186:
```suggestion
The boilerplate steps are generated automatically for you by the function generator
```

docs/serverlessFunctions.md
```diff
@@ -19,92 +11,622 @@ Creating serverless functions is easy with Redwood's function generator:
...
 
+    expect(result.statusCode).toBe(200)
+    expect(body.message).toContain('=')
+    expect(body.quotient).toEqual(4)
+  })
+```
+
+Then we can also add a test to handle the error when we don't provide a dividend:
 
-### Authentication
```
dac09, line 59: @dthyresson I removed this section until we update it with accurate information later

docs/serverlessFunctions.md
```diff
@@ -19,92 +11,622 @@ Creating serverless functions is easy with Redwood's function generator:
...
+
+### How to Test Webhooks
+
+[Webhooks](https://redwoodjs.com/docs/webhooks#webhooks) are specialized serverless functions that will verify a signature header to ensure you can trust the incoming request and use the payload with confidence.
+
+> **Note:** Want to learn more about webhooks? See a [Detailed discussion of webhooks](https://redwoodjs.com/docs/webhooks) to find out how webhooks can give your app the power to create complex workflows, build one-to-one automation, and sync data between apps.
+
+In the following example, we'll have the webhook interact with our app's database, so we can see how we can use **scenario testing** to create data that the handler can access and modify.
+
+>Why testing webhooks is hard
```
dac09, line 296:
```suggestion
> **Why testing webhooks is hard**
```

dthyresson:
@dac09 Added "Add advanced function mocking with fixtures" -> "Using Test Fixtures" section.

```js
// api/src/functions/divide/divide.test.ts
import { missingDivisor } from './divide.fixtures'

describe('divide serverless function',  () => {
// ... other test cases

  it('requires a divisor', async () => {
    const result = await handler(missingDivisor())

    const body = result.body

    expect(result.statusCode).toBe(400)
    expect(body.message).toContain('Please specify both')
    expect(body.quotient).toBeUndefined
  })

// ...
})

```

The import

```js
import { missingDivisor } from './divide.fixtures'
```

looks a bit odd.  Maybe the pattern is just to name the file `fixtures.ts`?

```js
import { missingDivisor } from './fixtures'
```

?

thedavidprice:
@dthyresson are you feeling good about this one being ready to merge?

dthyresson:
> @dthyresson are you feeling good about this one being ready to merge?

If the code to support the testing and the nesting is in, then yes, I feel very good about this documentation.

thedavidprice:
So this PR #3207, correct?

dac09:
Yup this PR is good to go!

dac09 (approved):

Issue #776, opened 2021-08-13
Cells Documentation refers to predictcovid.com but visiting site issues security warnings


dthyresson:
See the Loading section of the Redwood cells docs:

https://redwoodjs.com/docs/cells#loading

> For a production example, navigate to predictcovid.com, the first site made with Redwood

![image](https://user-images.githubusercontent.com/1051633/129416639-3fbdaa64-b893-4384-9a32-c27ee9c139a1.png)

When visiting the site, my eero firewall won't let me visit due to securing issues:

![image](https://user-images.githubusercontent.com/1051633/129416666-6080e954-d47d-4380-af01-810a59b07941.png)

Need a new example or remove?

thedavidprice:
+1 remove

PR #786, opened 2021-08-26
Removes predict covid site from cell loading documentation

dthyresson:
Fixes https://github.com/redwoodjs/redwoodjs.com/issues/776

Removes reference to the site based on

image

### DIFF
--- a/docs/cells.md
+++ b/docs/cells.md
@@ -163,8 +163,6 @@ export const afterQuery = (data) => ({...data})
 
 If the request is in flight, a Cell renders `Loading`.
 
-For a production example, navigate to predictcovid.com, the first site made with Redwood. Usually, when you first navigate there, you'll see most of the dashboard spinning. Those are `Loading` components in action!
-
 When you're developing locally, you can catch your Cell waiting to hear back for a moment if set your speed in the Inspector's **Network** tab to something like "Slow 3G".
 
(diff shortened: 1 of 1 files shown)


dthyresson:
Still gives me the error today:

image

dthyresson:
I'll report that the site is ok I guess.

dthyresson:
> confirming this is now working for me on Safari and Chrome. Best guess is the cert expired and we visited before it was renewed automatically on Netlify.

Actually, the link is not ssl.

```
http://predictcovid.com/
```

So no cert. And If they are deployed on Netlify, they should get a free Let's Encrypt cert, so very odd.

lachlanjc:
Yep, just re-checked & SSL is fully configured on Netlify!

peterp:
I think what @dthyresson is saying is that the link in our documentation is `http`, and it should be `https`, so we need to change that. It's odd that you're not redirected automatically.

peterp:
I had actually also experienced this error a few weeks ago 🤷

lachlanjc:
> I think what @dthyresson is saying is that the link in our documentation is `http`, and it should be `https`, so we need to change that. It's odd that you're not redirected automatically.

Ah, great flag. I can't see that Netlify has controls for this, but I'd certainly like that to be the case!

sarthaktexas:
If you use Cloudflare as your DNS provider, you can get the http to https redirect to work. Not sure about other providers. @lachlanjc

PR #788, opened 2021-08-29
[v0.37.0] Update for new scenario syntax

callingmedic911:
Closes https://github.com/redwoodjs/redwood/issues/3288

PR #792, opened 2021-09-16
[v0.37] Document using existing data in scenarios

KrisCoulson:
This pull request is branched off of  #788 which contains additional testing documentation changes

This pull request updates documentation regarding how to use existing data within a scenario and is directly related to this pull request in the redwood repo https://github.com/redwoodjs/redwood/pull/3386

My documentation skills are pretty rusty so feel free to request/make changes.

Tag @cannikin

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -1415,6 +1415,100 @@ scenario('retrieves a comment with post', async (scenario) => {
 })
 ```
 
+####  Relationships with an existing record
+If your models have relationships to each other and you need to connect them to an existing record, using the normal object syntax is just not going to cut it. 
+
+One example: a blog **Comment** has a parent **Post**, and both **Comment** and **Post** need to belong to an **Author**. With object syntax, there is no way to access the `authorId` of the `Author` we just created. We could potentially hard code the `id`, but that is just asking for trouble.
+
+```javascript
+// unable to access authorId
+export const standard = defineScenario({
+  post: {
+    first: {
+      data: {
+        name: 'First Post',
+        author: { create: { name: 'Kris' }},
+        comments: {
+          create: [
+            {
+              name: 'First Comment',
+              body: 'String',
+              authorId: // We want a different author...
+            },
+            {
+              name: 'First Comment Response',
+              body: 'String',
+              authorId: // We want the same author as post author...,
+            },
+          ],
+        },
+      }
+    }),
+  },
+})
+```
+
+
+
+When these use cases happen you can connect an existing `scenario` record using the distinctive name key as a function that returns an object.
+
+```javascript
+export const standard = defineScenario({
+  author: { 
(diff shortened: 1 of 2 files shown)


cannikin (approved):

docs/testing.md
```diff
@@ -1355,6 +1381,134 @@ scenario('creates a second comment', async (scenario) => {
...
+    first: {
+      data: {
+        name: 'First Post',
+        author: { create: { name: 'Kris' }},
+        comments: {
+          create: [
+            {
+              name: 'First Comment',
+              body: 'String',
+              authorId: // We want a different author...
```
jtoar, line 1436:
```suggestion
              authorId: // Here we want a different author...
```

jtoar (commented):
Hey @KrisCoulson thanks for adding this feature and the docs! Left a few comments let me know what you think

KrisCoulson:
@jtoar sounds good will take a look and make some tweaks

callingmedic911:
@KrisCoulson Hi! There are a few changes on the base PR. I think you might want to rebase. :)

KrisCoulson:
@callingmedic911 thanks for the update I rebase and committed your suggestions @jtoar should be good to go now with your review

docs/testing.md
```diff
@@ -1189,29 +1189,39 @@ When you use any of the generators that create a service (scaffold, sdl or servi
...
+        name: 'String',
+      },
     }
   },
 })
 ```
 
 This scenario creates two user records. The generator can't determine the intent of your fields, it can only tell the datatypes, so strings get prefilled with just 'String'. What's up with the `one` and `two` keys? Those are friendly names you can use to reference your scenario data in your test.
 
+The `data` key is one of Prisma's [create options](https://www.prisma.io/docs/reference/api-reference/prisma-client-reference#create). It's the same as in your Services—everything in the `one` and `two` keys actually just gets passed to Prisma's create. You can even create [relationships](https://redwoodjs.com/docs/testing.html#relationships) if you want.
```
Tobbe, line 1207: What do you mean "Services-everything"? It also feels like there's a missing word somewhere in this sentence. Like, should it be "... Services-everything in *that* the the `one`..."?

jtoar: The em dash there isn't rendering well in code. It'll look longer on the actual site, like "Services—everything". We could use a colon instead of that's better @Tobbe? 

I don't feel the need for an extra word in the second sentence (I'm reading it out loud to really try to see) but happy to restructure it another way to make it work

Tobbe: An em dash! That changes everything. I read it as a hyphen and didn't understand the sentence at all!. Indeed, no extra word needed now that I know it's an em dash.

jtoar (approved):

PR #793, opened 2021-09-16
Adds generator generator docs

cannikin:
Make live after https://github.com/redwoodjs/redwood/pull/3384 is merged.


docs/cliCommands.md
```diff
@@ -609,6 +609,44 @@ $ /redwood-app/node_modules/.bin/dev-server
 17:21:49 api | ► http://localhost:8911/user/
 ```
 
+### generate generator
+
+Copies a given generators template files to your local app for customization. The next time you generate that type again, it will use your custom template instead of Redwood's default.
```
jtoar, line 614: - add apostrophe
```suggestion
Copies a given generator's template files to your local app for customization. The next time you generate that type again, it will use your custom template instead of Redwood's default.
```

docs/cliCommands.md
```diff
@@ -609,6 +609,44 @@ $ /redwood-app/node_modules/.bin/dev-server
...
+
+If you wanted to customize the page generator template, run the command:
+
+```
+yarn rw generate generator page
+```
+
+And then check `web/generators/page` for the page, storybook and test template files. You don't need to keep all of these templates, you could customize just `page.tsx.template` and delete the others and they would still be generated, but using the default Redwood templates.
+
+The only exception to this rule is the scaffold templates. You'll get four directories, `assets`, `components`, `layouts` and `pages`. If you want to customize any one of the templates in those directories, you will need to keep all the other files inside of that same directory, even if you make no changes. (This is due to the way the scaffold looks up its template files.) For exmaple, if you wanted to customize only the index page of the scaffold (the one that lists all available records in the database) you would edit `web/generators/scaffold/pages/NamesPage.tsx.template` and keep the other pages in that directory. You _could_ delete the other three directories (`assets`, 'components`, `layouts`) if you just want to use the default templates for those.
```
jtoar, line 635: - typos
```suggestion
The only exception to this rule is the scaffold templates. You'll get four directories, `assets`, `components`, `layouts` and `pages`. If you want to customize any one of the templates in those directories, you will need to keep all the other files inside of that same directory, even if you make no changes. (This is due to the way the scaffold looks up its template files.) For example, if you wanted to customize only the index page of the scaffold (the one that lists all available records in the database) you would edit `web/generators/scaffold/pages/NamesPage.tsx.template` and keep the other pages in that directory. You _could_ delete the other three directories (`assets`, `components`, `layouts`) if you just want to use the default templates for those.
```

docs/cliCommands.md
```diff
@@ -609,6 +609,44 @@ $ /redwood-app/node_modules/.bin/dev-server
...
+
+**Usage**
+
+If you wanted to customize the page generator template, run the command:
+
+```
+yarn rw generate generator page
+```
+
+And then check `web/generators/page` for the page, storybook and test template files. You don't need to keep all of these templates, you could customize just `page.tsx.template` and delete the others and they would still be generated, but using the default Redwood templates.
```
jtoar, line 633: - feel like a semicolon or em dash is better here
```suggestion
And then check `web/generators/page` for the page, storybook and test template files. You don't need to keep all of these templates—you could customize just `page.tsx.template` and delete the others and they would still be generated, but using the default Redwood templates.
```

jtoar (commented):
Hey @cannikin, just caught a few typos but otherwise looks great!

docs/cliCommands.md
cannikin: image

cannikin:
Thanks @jtoar!

jtoar (approved):

PR #794, opened 2021-09-17
[Storybook] clarify that config files have to be made

jtoar:
Closes #3339 by making the fact that config files (and possible the config directory) have to be created. The "Getting Starting with Storybook" section is intentionally bare because I just don't have time to write an opinionated guide on it yet, but would like to at least include how to start it.

I've removed links to the code. In general I really like linking to code because I think it makes getting into the codebase less foreboding when you know where to go. But it results in dead links fairly often and I'm not updating them fast enough.

- https://github.com/redwoodjs/redwood/issues/3339

### DIFF
--- a/docs/storybook.md
+++ b/docs/storybook.md
@@ -21,6 +21,8 @@ This spins up Storybook on port `7910`.
 
 ## Configuring Storybook
 
+You only have to configure Storybook if you want to extend Redwood's default configuration, which handles things like how to find stories, configuring Webpack, startring Mock Service Worker, etc.
+
 There's two files you can add to your project's `web/config` directory to configure Storybook: `storybook.config.js` and `storybook.preview.js`. Note that you may have to create the `web/config` directory:
 
 ```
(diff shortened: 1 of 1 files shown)


jtoar:
@callingmedic911 let me know what you think!

callingmedic911:
Looks great! Only thing I'd suggest is to move up explanation that - we don't really need to use these two configuration file, it's just in case we want extend something - from `Configuring the Server with storybook.config.js` to `Configuring Storybook`. What do you think?

callingmedic911:
Also, do you think we should keep Github source links to main config and preview files? It could be confusing for new user but on the other hand, helpful if someone wants sneak peak what Redwood configures by default?

jtoar:
@callingmedic911 I applied both of your suggestions. I do like linking to the source, and maybe as we approach v1 I can be more confident in doing that since things shouldn't be changing as much. I'll also use tags in the links when we get to v1.

PR #808, opened 2021-10-01
Don't use pre-release version for latest release

callingmedic911:
Uses `getLatestRelease` from Octokit than `listReleases` & sorting externally, which fetches the "most recent non-prerelease, non-draft release, sorted by the created_at attribute.". Docs: https://octokit.github.io/rest.js/v18#repos-get-latest-release

The major difference now being: 
>The created_at attribute is the date of the commit used for the release, and not the date when the release was drafted or published.

Which should be okay for our case.
_____
@thedavidprice @doesnotexist @simoncrypta I'd be glad to have you people review this PR. :)

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -213,14 +213,12 @@ const get = async (pathToFile) => {
 
 // gets latest released version from GitHub
 const getLatestVersion = async () => {
-  const response = await octokit.repos.listReleases({
+  const latestRelease = await octokit.repos.getLatestRelease({
     owner: 'redwoodjs',
     repo: 'redwood',
   })
 
-  const sortedReleases = response.data.sort((a, b) => new Date(b.created_at) - new Date(a.created_at))
-
-  const version = sortedReleases[0].tag_name
+  const version = latestRelease.data.tag_name
   console.info(`For version ${version}`)
   return version
 }


doesnotexist (approved):
Nice, getLatestRelease function/endpoint seems to fit what we want.

simoncrypta (approved):
I was not sure if prerelease will be on `getLatestRelease`
but after reread the doc that make total sense that is what we need!
> The latest release is the most recent non-prerelease, non-draft release, sorted by the created_at attribute.

 will probably get a faster build time 😉
Thank you for this !

thedavidprice:
Looks good! Just running one more preview deploy. Thank you so much!

doesnotexist:
Just noticed that redwoodjs.com is showing v0.37.0 but v0.37.2 is the latest release on https://github.com/redwoodjs/redwood/releases
Is this a bug?

simoncrypta:
> Is this a bug?

Since the change is made at build time and no build was made since 3 days is normal I guess. It can be nice to plug the Netlify webhook to a GitHub Action of the RedwoodJS Framework to rebuild when a new release is publish.

thedavidprice:
That's actually another bug/issue!

Anytime a commit is merged into `main` on the Framework repo, Netlify is _supposed_ to trigger a rebuild of the .com site, which should update the version number. The Workflow is over here:
- https://github.com/redwoodjs/redwood/blob/main/.github/workflows/netlify-build.yaml

We need this for a couple .md files as well like the README.md (including all contributors).

BUT it doesn't work and I just have to manually trigger a Netlify build.

#help-wanted

PR #809, opened 2021-10-01
Fix: Typos in directive doc

callingmedic911:
**Me to typos:**

You shall not pass

cc: @thedavidprice @jtoar


dthyresson (approved):
Thanks @callingmedic911 ! I thought I ran these through a spell-check, but guess not.

PR #813, opened 2021-10-01
fix a few more typos

AntonioMeireles:
Hey all!

All the best &&  great weekend too!

António


jtoar:
Thanks @AntonioMeireles! Always a pleasure merging these 🚀

PR #816, opened 2021-10-03
doc: add Shadowenv setup command

virtuoushub:
See: https://github.com/redwoodjs/redwood/pull/3499

### DIFF
--- a/docs/cliCommands.md
+++ b/docs/cliCommands.md
@@ -1449,6 +1449,7 @@ yarn redwood setup <command>
 | `deploy`           | Setup a deployment configuration for a provider                                           |
 | `generator`        | Copy default Redwood generator templates locally for customization |
 | `i18n`             | Setup i18n                                                                                |
+| `shadowenv`        | Setup shadowenv                                                                           |
 | `tailwind`         | Setup tailwindcss and PostCSS                                                             |
 | `webpack`          | Setup webpack config file in your project so you can add custom config                    |
 | `tsconfig`         | Add relevant tsconfig, so you can start using TypeScript                                  |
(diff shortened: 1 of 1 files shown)


callingmedic911 (commented):
Nice! I have just added this suggestion to explain potential benefit of using Shadowenv. :)

Tobbe:
As a recent switcher from Windows to Mac, can I ask what the benefit of using shadowenv compared to just defining an alias is in the RW case?

callingmedic911:
> As a recent switcher from Windows to Mac, can I ask what the benefit of using shadowenv compared to just defining an alias is in the RW case?

I'll defer to @virtuoushub for the details but what I gathered is Shadowenv offers aliasing specific to the `cd`ed directory and cleans up after. Which means, you can have multiple RW projects and `rw` command will correspond to the one you're inside.

Tobbe:
> But I think you meant `alias rw="yarn rw"`

Yes, exactly :)  Sorry, I should have specified!

virtuoushub:
> As a recent switcher from Windows to Mac, can I ask what the benefit of using shadowenv compared to just defining an alias is in the RW case?

@Tobbe (cc @callingmedic911 ) Please see: https://github.com/redwoodjs/redwood/pull/1773#issuecomment-789261636 and the comments before for context around the choice of tooling.

thedavidprice:
> But for me, if I just continue using my alias rw="yarn rw", (and alias rwfw="yarn rwfw") what am I missing out on?

@Tobbe Understood. For those who know their shell, that would be simpler. I need to run through the process fully, but this is meant to be a solution that's cross-shell/platform. There may be more steps than I originally understood when this was proposed (way) back in the day.

Lastly, do check out the full documentation for shadowenv. I'm personally very interested in using it to "shadow" env vars per project.

virtuoushub:
Closing this as corresponding PR is also closed: https://github.com/redwoodjs/redwood/pull/3499#issuecomment-966724184

Issue #812, opened 2021-10-01
RBAC Cookbook needs to be updated to use Auth Validation Directives


dthyresson:
See: https://redwoodjs.com/cookbook/role-based-access-control-rbac

For v0.37, the way roles re checked has moved into SDLs.

All the examples here need updating.

PR #828, opened 2021-10-11
Update RBAC cookbook with example of how to protect service with GraphQL directives

jacebenson:
This should solve for #812 .

I updated the How to protect a service section with a SDL a bit of copy before it and then highlighted the updated lines at the bottom.

Here's what it looked like locally.
image

Let me know if there's more I should do

### DIFF
--- a/cookbook/Role-based_Access_Control.md
+++ b/cookbook/Role-based_Access_Control.md
@@ -510,23 +510,51 @@ export const requireAuth = ({ roles } = {}) => {
 
 #### How to Protect a Service
 
-```js
-import { db } from 'src/lib/db'
-import { requireAuth } from 'src/lib/auth'
+Services are protected by declaring a directive on the SDL.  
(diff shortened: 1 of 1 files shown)


jacebenson:
Let me know if that changes i made covered the concerns you had @dthyresson

dthyresson (changes requested):
Actually, I think that both techniques apply:

* Protect a service used as a resolver via the GraphQL dierctive
* Protect a service used elsewhere via the current described auth method
* and third, the forthcoming info here on protecting serverless functions https://github.com/redwoodjs/redwoodjs.com/pull/928
*

dthyresson:
@jacebenson i think RBAC has been updated to include this info perhaps? If so we can close

jacebenson:
the SDL isn't there anymore so it's not wrong.  I'm okay with it.

PR #845, opened 2021-10-18
userEvent instead of user

guledali:


cannikin:
It's been a while since I wrote this! Remind me how come we import `userEvent` but it doesn't actually appear in the tests anywhere? Won't ESLint complain about that?

jtoar (approved):
@cannikin I think @guledali's fix here just corrects for the fact that we import it as `user` but reference it in the subsequent codeblocks as `userEvent`

cannikin (approved):

PR #868, opened 2021-11-02
doc: Fix Email & Password Auth in Firebase options

usman-coe:
Email & Password Auth in Firebase options should be email and password, not username and password


jtoar (approved):
Thanks @usman-coe!

PR #870, opened 2021-11-04
Router doc improvements

mojombo:
Just a few improvements to some areas of the router docs that I thought could use some polish.

### DIFF
--- a/docs/router.md
+++ b/docs/router.md
@@ -406,7 +406,7 @@ In the following example, `/about/` will _not_ match `/about` and you will be se
 
 ## useParams
 
-Sometimes it's convenient to receive route parameters as the props to the Page, but in the case where a deeply nested component needs access to the route parameters, it quickly becomes tedious to pass those props through every intervening component. the router solves this with the `useParams` hook:
+Sometimes it's convenient to receive route parameters as the props to the Page, but in the case where a deeply nested component needs access to the route parameters, it quickly becomes tedious to pass those props through every intervening component. The router solves this with the `useParams` hook:
 
 ```js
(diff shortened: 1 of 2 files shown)


docs/router.md
```diff
@@ -375,28 +382,31 @@ Here we've created a custom `slug` route parameter type. It is defined by a `con
...
+  <Route path="/" page={HomePage} name="home" />
+  <Route path="/about" page={AboutPage} name="about" />
+  <Route notfound page={NotFoundPage} />
+</Router>
 ```
+
 ## useParams
 
-Sometimes it's convenient to receive route parameters as the props to the Page, but in the case where a deeply nested component needs access to the route parameters, it quickly becomes tedious to pass those props through every intervening component. RR solves this with the `useParams` hook:
+Sometimes it's convenient to receive route parameters as the props to the Page, but in the case where a deeply nested component needs access to the route parameters, it quickly becomes tedious to pass those props through every intervening component. the router solves this with the `useParams` hook:
```
jtoar, line 409: - capitalize t in the
```suggestion
Sometimes it's convenient to receive route parameters as the props to the Page, but in the case where a deeply nested component needs access to the route parameters, it quickly becomes tedious to pass those props through every intervening component. The router solves this with the `useParams` hook:
```

jtoar (commented):
Looks good to me @mojombo; the syntax highlighting on the code blocks makes those a lot more readable now!

Just two suggestions:
- I found one "t" that should be capitalized (see my comment)
- the URL has changed from `docs/redwood-router` to `docs/router`. We should probably add a redirect from `docs/redwood-router` right?

PR #871, opened 2021-11-04
update config defaults

thedavidprice:
Removing the copy/past from config.ts as it's incomplete (and therefore confusing).

Correcting defaults for new API endpoints.

### DIFF
--- a/docs/appConfiguration.md
+++ b/docs/appConfiguration.md
@@ -38,19 +38,19 @@ Error: Could not find a "redwood.toml" file, are you sure you're in a Redwood pr
 
 Configuration for the web side.
 
-| Key                           | Description                                                                 | Default                         | Context       |
-|:------------------------------|:----------------------------------------------------------------------------|:--------------------------------|:--------------|
-| `title`                       | Title of your Redwood App                                                   | `'Redwood App'`                 | `both`        |
-| `host`                        | Hostname to listen on                                                       | `'localhost'`                   | `development` |
-| `port`                        | Port to listen on                                                           | `8910`                          | `development` |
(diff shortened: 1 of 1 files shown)


docs/appConfiguration.md
```diff
@@ -72,19 +38,19 @@ Error: Could not find a "redwood.toml" file, are you sure you're in a Redwood pr
...
+| `host`                        | Hostname to listen on                                                       | `'localhost'`                   | `development` |
+| `port`                        | Port to listen on                                                           | `8910`                          | `development` |
+| `path`                        | Path to the web side                                                        | `'./web'`                       | `both`        |
+| `target`                      | Target for the web side                                                     | `'browser'`                     | `both`        |
+| `apiUrl`                      | Specify the URL to your api-server. Can be an absolute path or FQDN         | `'/.redwood/functions'`         | `production`  |
+| `apiGraphQLUrl`               | Optional: FQDN or absolute path to GraphQL function, without trailing slash | `'/.redwood/functions/graphql'` | `production`  |
+| `apiDbAuthUrl`                | Optional: QDN or absolute path to DbAuth function, without trailing slash   | `'/.redwood/functions/auth'`    | `production`  |
```
thedavidprice, line 50: @jtoar technically these should be something like `${apiUrl}/graphql` and `${apiUrl}/auth`. Do you think it's more clear and accurate if I were to use these template literals instead of defaults above?

jtoar: It's definitely more accurate. I feel like we can assume our readers can understand that syntax since this is a full stack JS framework. So go for it!

jtoar (approved):
See my one comment @thedavidprice but otherwise looks good!

PR #877, opened 2021-11-13
Fix module path for web testing utilities

callingmedic911:
Renames `@redwoodjs/testing` to `@redwoodjs/testing/web` for web side utility imports.


dac09 (approved):
This is correct @callingmedic911 - but just for your knowledge I also added a module alias in the jest config, so that even if you import from `rwjs/testing` it'll alias it to web (on the web side)

PR #880, opened 2021-11-14
Fixed CLI option for "allow host" for webpack 5

zakmandhro:
The documentation for allowed all hosts did not match webpack 5. New webpack option is: `--allowed-hosts all`. Updated accordingly.

Also added text to explain that this option is needed for SSH tunneling, such as ngrok or localhost.run.


callingmedic911 (approved):
Thanks @zakmandhro! Just tried with ngrok. Works well. 👍

PR #886, opened 2021-11-23
Fix docs for `excludeOperations` logger option

callingmedic911:
Thanks to @KrisCoulson: https://github.com/redwoodjs/redwoodjs.com/pull/863#discussion_r754798353.

Like Kris suggested I've updated the docs, and also moved it to the GraphQL page instead of where other Graphql logging options docs are.

### DIFF
--- a/docs/graphql.md
+++ b/docs/graphql.md
@@ -582,6 +582,43 @@ export const handler = createGraphQLHandler({
 })
 ```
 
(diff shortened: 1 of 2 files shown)


KrisCoulson (approved):
Looks good!

thedavidprice:
Thanks, all!

PR #895, opened 2021-11-29
Add section about babel configuration

dac09:
As title.

Other files were generated when I ran `yarn build`


jtoar:
@dac09 thanks for getting this started; per conversation with @thedavidprice we're going to give babel it's own section in the docs

dac09:
:+1:

jtoar (approved):
There's a lot more we could add here but gets the most important job (how to configure) done for now.

PR #921, opened 2021-12-02
Add instructions on how to remove Auth to the docs

edapm:
I have added instructions on how to remove authentication from a Redwood project. Feel free to suggest changes, this is just a first draft!

### DIFF
--- a/docs/authentication.md
+++ b/docs/authentication.md
@@ -1534,6 +1534,14 @@ const Routes = () => {
 }
 ```
 
+## Removing Authentication
+
+If you want to remove authentication, you will need to remove and change some files.
+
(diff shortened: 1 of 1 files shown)


dthyresson:
@dac09 I don't think these instructions are correct anymore given the need for `skipAuth()` directives.

I'm not sure this is even something one would want to document.

@edapm Could you describe your use case more? Could you just not enable authentication rather than removing it from your application? From a framework perspective, Redwood wants to highlight that security and auth is a priority and a core ability.

dthyresson (changes requested):
Could you describe why you might want to remove vs just disable (or not enforce auth via `skipAuth()`) in your application?

dthyresson:
@jtoar Can we get a triage and decision if we want to prioritize this PR? Thanks.

edapm:
@dthyresson @jtoar 
> @edapm Could you describe your use case more?
I used it to reset auth when I wanted to change Auth provider. Maybe there's a better way we could word this?

jtoar:
Hey @edapm thanks for contributing! One question to clarify the motivation behind these changes:

> I used it to reset auth when I wanted to change Auth provider. Maybe there's a better way we could word this?

It looks like these instructions are for removing authentication _completely_, not just changing the auth provider. Did you do this so that you could implement your own auth strategy from scratch?

jtoar:
Hey @edapm I'm going to close this for now but I'm planning to rework the auth docs soon so I'll review this again when I do so.

PR #927, opened 2021-12-21
Document why currentUser might be null after mocking

callingmedic911:
Closes https://github.com/redwoodjs/redwood/issues/3912.

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -620,7 +620,7 @@ Here we call `mockCurrentUser()` before the `render()` call. Right now our code 
 
 We introduced `waitFor()` which waits for a render update before passing/failing the expectation. Although `findByRole()` will wait for an update, it will raise an error if the element is not found (similar to `getByRole()`). So here we had to switch to `queryByRole()`, but that version isn't async, so we added `waitFor()` to get the async behavior back.
 
-Note that even after setting with `mockCurrentUser()` the initial render of the component might have still have null `currentUser` from `useAuth()`. It is updated after `getCurrentUser()` with mock data has finished execution.
+Note that even after setting with `mockCurrentUser()`, the initial render of the component might return `null` for the `currentUser` that you destructured from `useAuth()`. Once `getCurrentUser()` is executed, `currentUser` will be updated with the expected data.
 
 > Figuring out which assertions need to be async and which ones don't can be frustrating, we know. If you get a failing test when using `screen` you'll see the output of the DOM dumped along with the failure message, which helps find what went wrong. You can see exactly what the test saw (or didn't see) in the DOM at the time of the failure.
 >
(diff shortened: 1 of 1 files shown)


docs/testing.md
```diff
@@ -620,6 +620,8 @@ Here we call `mockCurrentUser()` before the `render()` call. Right now our code
 
 We introduced `waitFor()` which waits for a render update before passing/failing the expectation. Although `findByRole()` will wait for an update, it will raise an error if the element is not found (similar to `getByRole()`). So here we had to switch to `queryByRole()`, but that version isn't async, so we added `waitFor()` to get the async behavior back.
 
+Note that even after setting with `mockCurrentUser()` the initial render of the component might have still have null `currentUser` from `useAuth()`. It is updated after `getCurrentUser()` with mock data has finished execution.
```
cannikin, line 623: I wanted to just reword this a bit:

```suggestion
Note that even after setting with `mockCurrentUser()`, the initial render of the component might return `null` for the `currentUser` that you destructured from `useAuth()`. Once `getCurrentUser()` is executed, `currentUser` will be updated with the expected data.
```

cannikin (changes requested):

cannikin:
What do you think about this update @jtoar? I'm still not 100% sure I understand the problem: I haven't run into this issue, but everyone else says it's potentially confusing!

jtoar:
@cannikin I agree that it seems like if you use `waitFor` or an async matcher (not sure if that's the right term—`queryByRole`, `findByText`, etc), this just takes care of itself. 

But maybe where the confusion occurs is: it doesn't seem like we say that `waitFor` and/or an async matcher is necessary and if it is (seems like it is), why. Right now we say we use `waitFor` because `queryByRole` isn't async, but we don't explain why being async matters. So maybe I can work @callingmedic911's new paragraph into the previous one or just add a callout that says "hey, we're using waitFor and/or an async matcher because we need to wait for an update because currentUser is actually `null` in the first render..."

docs/testing.md
```diff
@@ -620,6 +620,8 @@ Here we call `mockCurrentUser()` before the `render()` call. Right now our code
 
 We introduced `waitFor()` which waits for a render update before passing/failing the expectation. Although `findByRole()` will wait for an update, it will raise an error if the element is not found (similar to `getByRole()`). So here we had to switch to `queryByRole()`, but that version isn't async, so we added `waitFor()` to get the async behavior back.
 
+Note that even after setting with `mockCurrentUser()`, the initial render of the component might return `null` for the `currentUser` that you destructured from `useAuth()`. Once `getCurrentUser()` is executed, `currentUser` will be updated with the expected data.
```
jtoar, line 623: @callingmedic911 what do you think of this?
```suggestion
The async behavior here is important. Even after setting the user with `mockCurrentUser()`, `currentUser` may be `null` during the initial render because it's being resolved. Waiting for a render update before passing/failing the exception gives the resolver a chance to execute and populate `currentUser`.
```

callingmedic911: This is way better! Thanks, @jtoar. Why is so hard for me to write I'm thinking. Hah!

jtoar (approved):

PR #928, opened 2021-12-22
Docs: How to authenticate users in serverless functions

dthyresson:
Documents how to authenticate in serverless function using the new `requireAuth` wrapper.

This feature was added in https://github.com/redwoodjs/redwood/pull/3785

### DIFF
--- a/docs/serverlessFunctions.md
+++ b/docs/serverlessFunctions.md
@@ -721,20 +721,18 @@ If your function receives an incoming Webhook from a third party, see [Webhooks]
 
 Serverless functions can use the same user-authentication strategy used by [GraphQL Directives](/docs/graphql#secure-directives) to [secure your services](/docs/graphql#secure-services) via the `useRequireAuth` wrapper.
 
-With the `useRequireAuth` wrapper, your serverless function can use the same `currentUser` and role checks defined in your application's `src/lib/auth.{js,ts}`.
-
 > If you need to protect an endpoint via authentication that isn't user-based, you should consider using [Webhooks](/webhooks) with a signed payload and verifier.
 
 #### How to Secure a Function with Redwood Auth
 
-The `useRequireAuth` wrapper preps your context so that you can **use** any of the `requireAuth`-related authetnication helpers in your serverless function:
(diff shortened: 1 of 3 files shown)


docs/serverlessFunctions.md
```diff
@@ -726,6 +717,80 @@ And, in some other cases, you may even want to limit how often the function is c
...
 If your function receives an incoming Webhook from a third party, see [Webhooks](https://redwoodjs.com/docs/webhooks) in the RedwoodJS documentation to verify and trust its payload.
 
+### Serverless Functions with User Authentication
+
+> Note that if you intend to implement a feature that requires user authentication, then using GraphQL, auth directives, and services is the preferred approach.
+
+Serverless functions can use the same user-authentication strategy used by [GraphQL Directives](/docs/graphql#secure-directives) to [secure your services](/docs/graphql#secure-services) via the `useRequireAuth` wrapper. 
+With the `useRequireAuth` wrapper, your serverless function can use the same `currentUser` and role checks defined in your application's `src/lib/auth.{js,ts}`.
+
+> If you need to protect an endpoint via authentication that isn't user-based, you should consider using [Webhooks](/webhooks) with a signed payload and verifier.
```
jtoar, line 727: @dthyresson previously this sentence was: "In addition, if you need to protect an endpoint that does not require not user-based authentication, you should consider using [Webhooks](/webhooks) with a signed payload and verifier."

That had too many negatives (does not require not...) so I edited it, but did I keep the logic the same?

jtoar (changes requested):
Hey @dthyresson I edited some of the copy and have a few questions I've left as comments.

One thing that wasn't obvious to me: after wrapping a serverless function handler in `useRequireAuth`, are you supposed to use `requireAuth` in the body of the function? Or is it already setup to require auth?

dthyresson:
> fter wrapping a serverless function handler in `useRequireAuth`, are you supposed to use `requireAuth` in the body of the function? Or is it already setup to require auth?

Good point. It means you can use the `auth` functions like `isAuthenticated()` or `hasRole()` or `requireAuth()` ... it lets you "use" any of the require auth-ivied functions.

jtoar (approved):

PR #945, opened 2022-01-17
Recipe: Sending Emails

Tobbe:
I wrote a cookbook recipe on building a newsletter signup form as a way to show how to send emails.

Do we want this as a recipe? Or should I just put in on my own blog and link it in the forums instead? I noticed there was an issue for creating an email recipe, but this isn't *exactly* what's described in that issue. https://github.com/redwoodjs/redwoodjs.com/issues/94

So let me know how you want me to proceed with this

Issue #94, opened 2020-04-21
Cookbook: send an email


cannikin:
GraphQL call that triggers an email via SendGrid.

We could also use this as an example of using one service inside another. Have a `log` table that logs sends. GraphQL call goes to a custom `emails.js` service, which imports `logs.js` service and creates the log entry.

dthyresson:
Even fancier (and incorporating the cron background job recipe idea) would be to:

* generate email
* persist email in a message q (in this case a table not a Redis q or other q)
* background cron job call function to send unsent messages (even 10 mins or so for testing)
* logs message sent (or if mail api call failed ... "reschedule")

I've used [Mailgun to send](https://documentation.mailgun.com/en/latest/api-sending.html#sending) quite a bit and solid experience --  [sdk](https://www.npmjs.com/package/mailgun-js)

jtoar:
Another chance to showcase [repeater.dev](repeater.dev)? #110

cannikin:
I thought this particular recipe should have the fewest number of external dependencies possible, ideally just the external mail service. I was planning on a SendGrid integration that's triggered by a custom Service (not backed by the database). As mentioned in the description that service would call the `logs` service (which *is* backed by the database) and would showcase one service using another (something @mojombo has been wanting to demonstrate for a while).

The Background Job recipe could then maybe build off of this one by sending the emails asynchronously...

jtoar:
That sounds great, excited to add a bona fide example of a service using a service.

jtoar:
@cannikin This would be a great one to do given the recent Discord activity

Tobbe:
@cannikin I created a PR for a email recipe for our cookbook. https://github.com/redwoodjs/redwoodjs.com/pull/945

It's not exactly what's discussed here (it's not calling one service from another). TBH I don't fully understand the logging stuff here. Wouldn't you just use @dthyresson's Logger for logging? Possibly sending the logs to LogDNA or Logtail or something like that?

cannikin:
It was a bit of a contrived example, but the idea was to have an audit trail in the database so that, for example, users could have a viewable/searchable list of sent emails in the app itself, instead of a logging stream for developer debugging. You could go to a "History" tab or something and see a list of all the emails that the app sent to/from you.

It wasn't the original intention of the email cookbook to do that, but after talking with Tom we thought that creating an audit trail would be a reasonable example of having one service (email) call another (audit/history).

Tobbe:
Solved by #949

PR #949, opened 2022-01-21
Recipe: Sending emails and importing service

Tobbe:
This is an attempt at solving #94 and an alternative to #945. Let's compare and see which one would benefit our users the most 🙂


cannikin (approved):
Interesting! I like this one as well since it does address the service-within-a-service issue. Unfortunately, not many people will see it unless they want to send an email and find this cookbook. We should probably add a section in the Services page of the docs that talks about composing services to surface it to more folks.

Tobbe:
> it does address the service-within-a-service issue. Unfortunately, not many people will see it

Yeah, I know. I tried to make it a bit more visible by making it a heading that shows up in the RHS nav menu. I've now also added a blurb about it in the top paragraph. Hopefully those two things will make google pick up on it, so if someone does a google search for it they'll find it.

Tobbe:
Got this message from @callingmedic911 

> Both of them are great! If I have to select one, I'd say #949 because it focuses on only sending email and nothing more. With the newsletter example in #945, in comparison, people might expect the whole thing, how to schedule the newsletter, subscription and unsub management (even though you hinted in the last section that there's more to it).
Just my opinion. In any case, both are super helpful for the community.

Thanks to his message I've decided to move forward with this PR, and close the other one

Tobbe:
GitHub CDN FTW!

image

image

image

image

image

Issue #879, opened 2021-11-13
How to use GraphQL fragments with redwood


orta:
I didn't really find a nice place in the GraphQL document to add this, so I'm making this issue so it would turn up in search results. There are two things you need to do:

### Setup tooling

The GraphQL VS Code extension needs to know the fragment exists, so you need to change your `graphql.config.js` to start referring to documents:

```diff
const { getPaths } = require("@redwoodjs/internal")

module.exports = {
   schema: getPaths().generated.schema,
+  documents: 'packages/app/src/components/**/*.ts'
}
```

This means the extension will start looking for `gql` tags inside those `*.ts` files.

### Setup your fragments

Co-locate your fragment with your component by using an `export const` with a `gql` string:

```tsx
import { Link, routes } from "@redwoodjs/router"
import { PuzzleThumbnail, RenderThumbnailPropsHierarchy } from "./RenderPuzzleThumbnail"

export const thumbnailFragment = gql`
  fragment Thumbnail on Puzzle {
    id
    name
    gameID
    puzzle

    currentAccountGamePlayed {
      id
      boardState
    }

    game {
      id
      slug
      assetsSha
      thumbnailJSPath
    }
  }
`

export const UserPuzzleThumbnail = (props: { puzzle: RenderThumbnailPropsHierarchy } & { puzzle: { name?: string } }) => {
  const url = props.puzzle.currentAccountGamePlayed
    ? routes.playGameWithID({ gameSlug: props.puzzle.game.slug, gamePlayID: props.puzzle.currentAccountGamePlayed.id })
    : routes.playGame({ gameSlug: props.puzzle.game.slug })

  return (
    <div key={props.puzzle.id} className="m-4">
      <Link to={url}>
        <p>{props.puzzle.name || "No name"}</p>
        <div className="w-60 h-60">{props.puzzle.game.thumbnailJSPath && <PuzzleThumbnail {...props.puzzle} />}</div>
        <p>{props.puzzle.game.slug}</p>
      </Link>
    </div>
  )
}
```

Then in the closest query import the fragment and use it:

```ts
import { thumbnailFragment, UserPuzzleThumbnail } from "../UserRenderPuzzle"

export const QUERY = gql`
  query DashboardQuery {
    todaysRecommendation(category: "production") {
      id
      category
      day
      puzzles {
        puzzle {
          ...Thumbnail
        }
      }
    }
  }
  ${thumbnailFragment}
`
```

orta:
[...]

PR #952, opened 2022-01-24
Docs: How to use GraphQL fragments with redwood

dthyresson:
Closes https://github.com/redwoodjs/redwoodjs.com/issues/879

@orta included an example of how to use GraphQL fragments with Redwood in https://github.com/redwoodjs/redwoodjs.com/issues/879

I adapted his example to be closer to the test app and tutorial for reuse fields on a Post fragment.

I think I have the example correct -- it is working in a sample project with the code shows.

Few questions:

1. Still not 100% certain how the VS Code extension works. I had to change the `documents` setting from
```
+  documents: 'packages/app/src/components/**/*.ts'
```

to 

```
+  documents: 'packages/app/src/components/**/*.ts',
```

because I did not have my app setup as he has.

2. I get some sort of ts lint issue when using the fragment:

image

And I cannot figure out how to resolve.

### DIFF
--- a/docs/graphql.md
+++ b/docs/graphql.md
@@ -528,6 +528,186 @@ export const handler = createGraphQLHandler({
 
 > Note: Check-out the [in-depth look at Redwood Directives](https://www.redwoodjs.com/docs/directives) that explains how to generate directives so you may use them to validate access and transform the response.
 
+## Fragments
+
+[GraphQL Fragments](https://graphql.org/learn/queries/#fragments) let your components be more self-contained and simpler to refactor because they promote reuse by sharing fields between multiple queries and mutations.
+
+Let's say that everywhere you need to render a `Post`, you always need to show the same set of fields -- even nested fields, such as the `Author` name.
+
+To do that, you can declare a `PostDetails` fragment that can be used with any `Post` object:
+
+```graphql
+fragment PostDetails on Post {
+  id
+  title
+  body
+  author {
+    id
+    name
+  }
+}
+```
+
+You can now include the `PostDetails` fragment in any of your of queries and mutations that use `Post` objects, like:
+
+```graphql
+query FindPostQuery {
+  post(id: 1) {
+    ...PostDetails
+  }
+}
+```
+
+> You precede an included fragment with three periods (...), much like the spread syntax.
+
+Or, if you needed additional `Post` fields:
+
+```graphql
+query FindPostQuery {
+  post(id: 1) {
+    ...PostDetails
+    createdAt
+    updatedAt
+  }
+}
+```
+
+If you later change which fields are included in the `PostDetails` fragment, the up-to-date fields are automatically included in operations that use the fragment.
+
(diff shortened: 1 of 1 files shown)


jtoar:
@dthyresson Reviewing this now. One quick question: I don't see a difference between the two code blocks. Did you mean for the second one to be something like:
```diff
+  documents: 'web/src/components/**/*.ts',
```

dthyresson:
> @dthyresson Reviewing this now. One quick question: I don't see a difference between the two code blocks. Did you mean for the second one to be something like:
> 
> ```diff
> +  documents: 'web/src/components/**/*.ts',
> ```

Ah, it can be simply:

```
documents: 'web/src/components/**/*.ts',
```

dthyresson:
@orta I had a question about:

```
Still not 100% certain how the VS Code extension works. I had to change the documents setting from

+  documents: 'packages/app/src/components/**/*.ts'
to

+  documents: 'packages/app/src/components/**/*.ts',
```

I was not sure what this did and if the config is ok. 

My fragments were working with or without this addition.

Could it be that your fragments are located outside of `web` and if so then you have to configure graphql?

dthyresson:
Also @orta any idea why I have a TS lint issue as seen






and above?

orta:
Check how many copies of graphql-js there are in your tree, I have a feeling I had to use `resolutions` originally

orta:
```diff
+  documents: 'packages/app/src/components/**/*.ts'
```

Is wrong (I used `packages/app/src/components/**/*.tsx`) before:

Screen Shot 2022-01-26 at 5 01 10 PM

After:
Screen Shot 2022-01-26 at 5 06 07 PM

Basically, the graphql extension without this change _only_ uses the generated schema file to understand your schema, but as the fragments live outside of that the extension needs to know where to pick those up

dthyresson:
> fragments live outside of that the extension needs to know where to pick those up

Oh I see now how you use it. I was importing and also using a placeholder as per some of the Apollo fragment examples.

This looks so much nicer. I’ll try. Thanks!

orta:
On a blank project I didn't get the eslint errors?

Screen Shot 2022-01-26 at 5 11 47 PM

dthyresson:
@orta I must be doing something wrong because when I try to do just:

image

I get the error: `Error: Cannot read properties of undefined (reading 'kind')`

image


This is my `Post.tsx` component with the fragment

```ts
import { Link, routes } from '@redwoodjs/router'

export const PostDetailsFragment = gql`
  fragment PostDetails on Post {
    id
    title
    body
  }
`

export const Post = ({ post }) => {
  return (
    <div key={post.id}>
      <Link to={routes.post({ id: post.id })}>
        <h2>{post.title}</h2>
        <p>{post.body}</p>
      </Link>
    </div>
  )
}
```

orta:
I think you need to include the fragment inside the query, it's probably doing a lookup and not finding the fragment

dthyresson:
This should be a guide instead. Closing as still a few details to work out.

Issue #954, opened 2022-01-28
Update Docs for Render Deploy to match the cli


jacebenson:
I was wrking on trying out Render today and I was following the `yarn rw setup deploy` step for render regarding database selection.

I got the following error.
![image](https://user-images.githubusercontent.com/638764/151463095-dda1ac73-e8ff-4298-afbd-cc1db0982306.png)
```
Invalid values:
  Argument: database, Given: "postgres", Choices: "none", "postgresql", "sqlite"
```

The docs say you can use `postgres` here: https://redwoodjs.com/docs/deploy#render-tldr-deploy.  But the CLI says its needs to be postgresql

Should just be a typo fix: https://github.com/redwoodjs/redwoodjs.com/blob/main/docs/deploy.md?plain=1#L104

simoncrypta:
Hi @jacebenson ! Do you think we should update the doc, or we should update the CLI? I'm not the best in naming, but I'm always in the clan of shorter is better for CLI 😛. 

Also, I think you merged the PR into your own fork. Can you retry from your main branch to the redwood repo?

jacebenson:
Hey, @simoncrypta I'm not sure where the right place should be or if maybe we allow for both?  But it should be consistent.

I figured the least amount of work would be just to update the docs.

I'll retry my PR, I am not sure why or what I was doing.

If we need to update the CLI to match then the docs are right.

PR #955, opened 2022-01-28
Update Render Deploy option to match CLI command solves #954

jacebenson:


docs/deploy.md
```diff
@@ -101,7 +101,7 @@ Render is a unified cloud to build and run all your apps and websites with free
 If you simply want to experience the Render deployment process, including a Postgres or SQLite database, you can do the following:
 1. create a new redwood project: `yarn create redwood-app ./render-deploy`
 2. after your "render-deploy" project installation is complete, init git, commit, and add it as a new repo to GitHub or GitLab 
-3. run the command `yarn rw setup deploy render`, use the flag `--database` to select from `postgres`, `sqlite` or `none` to proceed without a database
+3. run the command `yarn rw setup deploy render`, use the flag `--database` to select from `postgresql`, `sqlite` or `none` to proceed without a database
```
simoncrypta, line 104: I would also add that postgresql is the default choice
```suggestion
3. run the command `yarn rw setup deploy render`, use the flag `--database` to select from `postgresql`, `sqlite` or `none` to proceed without a database [default : `postgresql`]
```

simoncrypta (changes requested):

simoncrypta (approved):
All good ! Thank you for the correction 😊

PR #962, opened 2022-02-03
Updated docs for new Serverless deploy

cannikin:


docs/deploy.md
```diff
@@ -171,9 +171,80 @@ From the Vercel Dashboard you can access the full settings and information for y
...
+
+```bash
+yarn rw deploy serverless --first-run
+```
+
+The first time you deploy your app we'll first deploy just the API side. Once it's live we can get the URL that it's been deployed to, and add that as an environment variable so that web side will know what it is during build-time (it needs to know where to send GraphQL and function requests).
+
+Half-way through the first deploy you'll be prompted if you want to add the API_URL to `.env.production` (which is similar to `.env` but is only used when `NODE_ENV=production`, like when building he web and api sides). Make sure you say `Y`es at this prompt and then it will continue to deploy the web side.
+
+Once that command completes you should see a message including the URL of your site! Open that URL and hopefully everything works as expected!
```
dac09, line 215:
```suggestion
Once that command completes you should see a message including the URL of your site! Open that URL and hopefully everything works as expected!

> **Heads up**
> If you're getting an error trying to load data from the API side, its possible you're still pointing at your local database.
> Remember to add a DATABASE_URL env var to your `.env.production` file that is created, pointing at the database you want to use on your deployed site. Since your stack is on AWS, RDS might be a good option, but you might find it easier/quicker to setup databases on other providers too, such as [Railway](https://railway.app/) or [Supabase](https://supabase.com/)

```

ajcwebdev:
Found an edge case, long project names will trigger a REGEX and fail the web build:

```
Resource handler returned message:

"Invalid request provided: 2 validation errors detected:

Value 'redwood-serverless-framework-web-production-us-east-1-web-response' at 'name' failed to satisfy constraint:
Member must satisfy regular expression pattern:
^[a-zA-Z0-9-_]{1,64}$;

Value 'redwood-serverless-framework-web-production-us-east-1-web-response' at 'name' failed to satisfy constraint:
Member must have length less than or equal to 64

(
  Service: CloudFront,
  Status Code: 400,
  Request ID: 0926135e-d48e-4cd8-b0d6-89d43f72c6a5,
  Extended Request ID: null
)"

(
  RequestToken: 1b75ed31-8efe-8bd8-ca76-ca92cc223c6e,
  HandlerErrorCode: InvalidRequest
)
```

dac09:
> Found an edge case, long project names will trigger a REGEX and fail the web build:

I don't believe this is an error coming from us, but a limitation with deploying with serverless, thankfully the error message is clear enough to act on.

ajcwebdev:
Yup, agreed it's an easy fix, I just modified the `service` names in each `serverless.yml` file and [redeployed](https://d1roiykqcb1rwu.cloudfront.net/). I feel like it would be possible to provide some kind of validation of the name length to keep the user from blowing themselves up, but I understand if that's not something we want to handle on our end. Might be worth at least including a warning note in the docs.

PR #964, opened 2022-02-03
Cells: Update section on props

Tobbe:
Update information on exactly what props are available to the different Cell sub-components.
Also add info on prop name clashes.

Related framework PR: https://github.com/redwoodjs/redwood/pull/4304


jtoar (approved):
This was awesome @Tobbe thanks!

PR #978, opened 2022-02-22
cli: Add docs for 'secret'

Tobbe:
We were missing docs for the `secret` cli command. So I went ahead and added that, together with documenting the "append SESSION_SECRET to your .env" use-case.

Related PR: https://github.com/redwoodjs/redwood/pull/4523


thedavidprice (changes requested):
Since this is a `generate` command, I think it should be after the `generate sdl` section here:
https://deploy-preview-978--redwoodjs.netlify.app/docs/cli-commands#generate-sdl

copy looks great = ✅

Tobbe:
> I think it should be after the `generate sdl` section here

Of course! 

Fixed.

thedavidprice (approved):

PR #982, opened 2022-02-28
Svix webhook docs

Tobbe:
Document changes introduced in https://github.com/redwoodjs/redwood/pull/4598


docs/webhooks.md
```diff
@@ -191,6 +198,97 @@ export const handler = async (event: APIGatewayEvent) => {
...
+import {
+  verifyEvent,
+  VerifyOptions,
+  WebhookVerificationError,
+} from '@redwoodjs/api/webhooks'
+
+import { logger } from 'src/lib/logger'
+
+export const handler = async (event: APIGatewayEvent) => {
+  const discourseInfo = { webhook: 'clerk' }
```
dthyresson, line 218: Maybe make this clerkInfo or userInfo -- instead of discourseInfo?

dthyresson (changes requested):
Just one variable name suggestion otherwise looks good!

docs/webhooks.md
Tobbe: Done. Went with `clerkInfo`

dthyresson (approved):

PR #985, opened 2022-03-02
Documents custom option for log formatter

dthyresson:
Adds documentation for the logFormatter custom option as implemented in https://github.com/redwoodjs/redwood/pull/4619


simoncrypta (approved):
LGTM 

I don't know if an image of the result can be relevant ?

PR #988, opened 2022-03-04
remove Serverless CI/CD copy

thedavidprice:
Redwood does not work with Serverless CI/CD. This removes the specific CI/CD instructions, highlights that it won't work, but adds reasons why it's still helpful.

### DIFF
--- a/docs/deploy.md
+++ b/docs/deploy.md
@@ -234,23 +234,15 @@ There are even more places you can get environment variables from, check out Ser
 
 ### Serverless Dashboard
 
+> **Note:**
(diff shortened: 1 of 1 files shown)


cannikin:
Updates look good!

But man I would have bet money that @dac09 and I saw this working...we were on a Tuple call working on the code and he went to their dashboard, played with some config and we were all excited it was working...was that just a dream I had? Maybe it wasn't for automated deploys but something else...

cannikin (approved):

thedavidprice:
@cannikin yeah, it's all so very close. I connected all the things, but the things don't actually connect and work for coordinated deploys. But the dashboard is still the way to go. Just not the CI/CD.

Good news is that Serverless wants to coordinate with us about next steps to make it work. TBD

PR #996, opened 2022-03-09
Document `nameVariants` provided to generator templates

realStandal:
This PR documents the [nameVariants](https://github.com/redwoodjs/redwood/blob/d48f4bb1ebf61b67da8137c0f87bec816dc7de75/packages/cli/src/lib/index.js#L44) which are provided to an application generator templates. I've had to search for these a few times when customizing my templates, and figured it'd be handy to have them easily accessible.


jtoar (approved):
Great idea @realStandal; everything looks good to me. Merging now 🚀

PR #998, opened 2022-03-10
describe how to use Redirect's options prop

odjhey:
Docs for this change https://github.com/redwoodjs/redwood/pull/4704.

@Tobbe , please check, i'm not confident with my english esp for docs related usages ( ｡･ω･｡)  so feel free to make big changes.

Thanks and good day!

### DIFF
--- a/docs/router.md
+++ b/docs/router.md
@@ -498,6 +498,8 @@ const SomePage = () => {
 }
 ```
 
+In similar fashion to `navigate`'s second argument: `navigate(_, { replace: true })`, we can *replace* the top item of the browser history stack (instead of pushing a new one). We achieve this by using `Redirect`'s `options` prop. Above example could be converted to `<Redirect to={routes.home()} options={{ replace: true }}/>` to have this effect.
+
 ## Code-splitting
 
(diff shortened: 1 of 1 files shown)


docs/router.md
```diff
@@ -498,6 +498,8 @@ const SomePage = () => {
 }
 ```
 
+In similar fashion to `navigate`'s second argument: `navigate(_, { replace: true })`, we can *replace* the top item of the browser history stack (instead of pushing a new one). We achieve this by using `Redirect`'s `options` prop. Above example could be converted to `<Redirect to={routes.home()} options={{ replace: true }}/>` to have this effect.
```
Tobbe, line 501:
```suggestion
In addition to the `to` prop, `<Redirect />` also takes an `options` prop. This is the same as [`navigate()`](https://redwoodjs.com/docs/router#navigate)'s second argument: `navigate(_, { replace: true })`. We can use it to *replace* the top item of the browser history stack (instead of pushing a new one). This is how you use it to have this effect: `<Redirect to={routes.home()} options={{ replace: true }}/>`.
```

odjhey:
thanks @Tobbe!

jtoar (approved):
Looks great @odjhey, thanks!

PR #1000, opened 2022-03-11
Escape '|' character to fix formatting on markdown table

nickpdemarco:
A one-character change to fix Markdown, and some whitespace to tidy it up. 

@reviewer - my vscode was making some whitespace changes on save which, I assume, are just based on my local formatting rules for markdown. I took a look at this repo's `prettier.config.js` and it doesn't seem to have Markdown-specific rules. Is there a way I can configure my environment to ensure I conform to Redwood's formatting preferences for `.md`?

### DIFF
--- a/docs/cells.md
+++ b/docs/cells.md
@@ -53,16 +53,16 @@ Off the bat, this file exports five constants: `QUERY`, `Loading` , `Empty` , `F
 
 With Cells, you have a total of seven exports to work with:
 
-| Name          | Type              | Description                                                  |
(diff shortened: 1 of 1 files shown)


jtoar:
Hey @nickpdemarco great question. I think the short answer is no, we don't have any, or if we do, it's all scattered/different. I use a vscode markdown extension to format the tables. But we should probably add config for markdown to the repo instead of putting the onus on you and other contributors to add it to your local config. I'll make that a todo!

jtoar (approved):

PR #1009, opened 2022-03-17
Add docs on Flightcontrol deploy

beerose:
Related issue: https://github.com/redwoodjs/redwood/pull/4788


thedavidprice:
moved over here with docs migration https://github.com/redwoodjs/redwood/pull/4826

cc @flybayer

PR #1023, opened 2022-08-14
Motion Framer added 

miguelgargallo:
🚨 HEADS UP

In preparation for the 1.0.0 release, the Redwood Docs have been integrated with the main Redwood Framework redwoodjs/redwood repo. We are in the final stages of the migration. Once complete, this repo will be archived.

> Please do not open new PRs here. Instead, open them at https://gi🥰thub.com/redwoodjs/redwood/pulls

The new doc site uses Docusaurus. The Docs are located here:
- https://github.com/redwoodjs/redwood/tree/main/docs

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,15019 @@
+{
+  "name": "redwoodjs.com",
+  "version": "0.0.1",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "redwoodjs.com",
+      "version": "0.0.1",
+      "dependencies": {
+        "@fullhuman/postcss-purgecss": "2.3.0",
+        "@octokit/rest": "16.43.2",
+        "algoliasearch": "4.11.0",
+        "autoprefixer": "9.8.8",
+        "blueimp-md5": "2.19.0",
+        "cameronjs-html-webpack-plugin": "0.5.1",
+        "clipboard": "2.0.8",
+        "dotenv": "8.6.0",
+        "framer-motion": "^7.1.0",
+        "highlight.js": "10.7.3",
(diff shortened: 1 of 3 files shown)


ahaywood:
I stumbled upon this PR because any storybook story with Framer Motion is breaking my Storybook build.

If I launch Storybook without the story and then add it, it doesn't have any trouble. But, if I create the story and then build it, it breaks.

This is the only other seemingly related issue I've been able to find so far: https://stackoverflow.com/questions/72710138/framer-motion-with-storybook-error-in-build-environment

Happy to provide additional information if needed.

miguelgargallo:
> I stumbled upon this PR because any storybook story with Framer Motion is breaking my Storybook build.
> 
> If I launch Storybook without the story and then add it, it doesn't have any trouble. But, if I create the story and then build it, it breaks.
> 
> This is the only other seemingly related issue I've been able to find so far: https://stackoverflow.com/questions/72710138/framer-motion-with-storybook-error-in-build-environment
> 
> Happy to provide additional information if needed.

I think I might find something useful, https://blog.sethcorker.com/scroll-animation-with-framer-motion/ from here's an example https://framer-motion-stories.sethcorker.com/?path=/story/framer-motion-scroll--letter

ahaywood:
Thanks for sharing that link... and it does prove that Storybook and framer-motion can co-exist, but I believe something else is going on... 

I found a couple of other issues with framer-motion: https://github.com/framer/motion/issues/1504 and https://github.com/framer/motion/issues/1447

In fact, locking into 6.2.4 resolved the issue for me, but that's a full version behind what's currently available. 😞

miguelgargallo:
Yeah! That's why I am currently integrating on it, it would be awesome some day to use it! Let's go RedwoodJS Team!!!

